PR #2, opened 2017-05-23
Remove reference to zap

glibsm:
If that particular benchmark was testing something in particular we
shuold bring it back, but not rely on another external package to test
something related to the config.

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -1,5 +1,5 @@
 hash: ca25df7c2a2283dbce17355e832e14051a54ac56c32fe81c0684321de31217ff
-updated: 2017-05-23T11:01:34.102390715-07:00
+updated: 2017-05-23T11:22:37.492884021-07:00
 imports:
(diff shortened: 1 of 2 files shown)


glide.lock
```diff
@@ -25,18 +25,7 @@ testImports:
   subpackages:
   - assert
   - require
-- name: go.uber.org/atomic
-  version: 4e336646b2ef9fc6e47be8e21594178f98e5ebcf
 - name: go.uber.org/tools
   version: ce2550dad7144b81ae2f67dc5e55597643f6902b
   subpackages:
   - update-license
-- name: go.uber.org/zap
```
glibsm: this lock change definitely looks like a win. We don't want to pull in zap for people who want to use the config package

alsamylkin: This was a E2E test, so in theory this could be a test import. But it is ok to remove it, because we have other tests for functions unmarshalling.

PR #3, opened 2017-05-23
Add travis, badges, slightly edit README

glibsm:


.travis.yml
```diff
@@ -0,0 +1,14 @@
+language: go
+
+go_import_path: go.uber.org/config
+
+go:
+  - 1.7
```
alsamylkin: Do we still need it?

breerly: Yup - Uber still supports 1.7 atm.

PR #6, opened 2017-05-23
Codecov

### LINKED ISSUE #5: codecov.io
Need to enable code coverage and I'd like to try out codecov.io to see how it compares with coveralls

glibsm:
Seeing if the super basic setup is enough. 

Fixes #5

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,15 @@
+coverage:
+  range: 50..95
+  round: down
+  precision: 2
+
+  status:
+    project:                   # measuring the overall project coverage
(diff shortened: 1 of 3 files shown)


breerly:
I *believe* you need a `.codecov.yaml` file... check out their docs for Go

glibsm:
I'm just trying to follow https://github.com/codecov/example-go to start with

breerly:
neat!

breerly:
We should set the threshold to 95% so the build fails if we fall below that number.

PR #12, opened 2017-05-25
Implement childKeys method

alsamylkin:
Simply populate a slice, map and then combine the results.

### DIFF
--- a/static_provider_test.go
+++ b/static_provider_test.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"sort"
 )
 
 func TestStaticProvider_Name(t *testing.T) {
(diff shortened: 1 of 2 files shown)


value.go
```diff
@@ -148,16 +148,20 @@ func (cv Value) WithDefault(value interface{}) Value {
...
 
-// TODO: Support enumerating child keys
-// 1. Add a method on Provider to get the child keys for a given prefix
-// 2. Implement in the various providers
-// 3. Merge the list here
-// 4. Return the set of keys.
-
-// ChildKeys returns the child keys
-// TODO(ai) what is this and do we need to keep it?
 func (cv Value) ChildKeys() []string {
```
anuptalwalkar: documentation?

value.go
```diff
@@ -148,16 +148,20 @@ func (cv Value) WithDefault(value interface{}) Value {
...
-	return nil
+	var res []string
+
+	// intentionally ignoring the error, because we can't do much about it
+	// and simply return an empty slice in that case.
+	cv.Populate(&res)
+
+	var m map[string]interface{}
+	cv.Populate(&m)
+	for k := range m {
```
anuptalwalkar: `for k, _ := range m` for readability niceness

value.go
```diff
@@ -148,16 +148,20 @@ func (cv Value) WithDefault(value interface{}) Value {
...
-
-// ChildKeys returns the child keys
-// TODO(ai) what is this and do we need to keep it?
 func (cv Value) ChildKeys() []string {
-	return nil
+	var res []string
+
+	// intentionally ignoring the error, because we can't do much about it
+	// and simply return an empty slice in that case.
+	cv.Populate(&res)
```
glibsm: i think you should check the error (to make errcheck happy) and if you incur it, just return an empty set or something? or nil

value.go
glibsm: that syntax is a warning

static_provider_test.go
```diff
@@ -271,3 +272,26 @@ func TestPopulateForMapsWithNotAssignableKeyTypes(t *testing.T) {
...
+	})
+
+	op := func(t *testing.T, key string, expected []string) {
+		keys := p.Get(key).ChildKeys()
+		sort.Strings(keys)
+		assert.Equal(t, expected, keys)
+	}
+
+	t.Run("Map", func(t *testing.T) { op(t, Root, []string{"one", "slice", "two"}) })
+	t.Run("Slice", func(t *testing.T) { op(t, "slice", []string{"13", "42"}) })
```
glibsm: I don't get this. 

Why are we returning array values as child keys? If it was `[]string{0,1}` I would sort of understand, but even then I'd say lets just do maps, no?

alsamylkin: Well, I was thinking about it as a collection of elements:
```yaml
- 13
- 42
```
but you right, the go way, would be {0, 1}

PR #13, opened 2017-05-26
Fix group provider for maps

alsamylkin:
Group provider iterates over all underlying providers until it finds a value, which works for structs, because we know key in advance, but fails for maps, where keys are dynamic.

We are going to reuse the mergeMaps function to merge values from all providers.

### DIFF
--- a/provider_group.go
+++ b/provider_group.go
@@ -28,34 +28,28 @@ type providerGroup struct {
 // NewProviderGroup creates a configuration provider from a group of backends.
 // The highest priority provider is the last.
 func NewProviderGroup(name string, providers ...Provider) Provider {
-	l := len(providers)
-	p := providerGroup{
+	return providerGroup{
 		name:      name,
-		providers: make([]Provider, l),
+		providers: providers,
 	}
-
-	for i := 0; i < l; i++ {
-		p.providers[i] = providers[l-i-1]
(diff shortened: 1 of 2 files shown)


provider_group.go
```diff
@@ -28,34 +28,28 @@ type providerGroup struct {
...
-	cv := NewValue(p, key, nil, false, GetType(nil), nil)
-
 	// loop through the providers and return the value defined by the highest priority provider
+	var res interface{}
+	found := false
 	for _, provider := range p.providers {
 		if val := provider.Get(key); val.HasValue() && !val.IsDefault() {
-			cv = val
-			break
+			res = mergeMaps(res, val.value)
```
anuptalwalkar: this mean if the value exists in yaml and dynamic providers both, we will use one from yaml?

anuptalwalkar: Oh I see you removed the reverse order. +1

alsamylkin: We save the value from the yaml first and then will overwrite it with the value from the dynamic provider.
This is covered in UT by the `name` key: the first provider has it set to `test` and second has `fx`.

alsamylkin: A provider doesn't know if a value is populating a map or a struct, so we have to iterate over all providers.

PR #14, opened 2017-05-26
Skip parallel load test

alsamylkin:
The loader logic is extremely convoluted and not actually concurrent safe, which makes this test to fail sometimes.

### DIFF
--- a/config_test.go
+++ b/config_test.go
@@ -27,10 +27,9 @@ import (
 	"io/ioutil"
 	"os"
 	"path"
+	"path/filepath"
 	"sync"
 	"testing"
-
-	"path/filepath"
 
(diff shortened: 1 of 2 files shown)


config_test.go
```diff
@@ -572,15 +574,17 @@ func TestParallelLoad(t *testing.T) {
 		assert.Equal(t, "bane", p.Get("vilain").String())
 	}
 
+	count := 100
 	wg := sync.WaitGroup{}
-	wg.Add(2)
```
alsamylkin: Having just 2 goroutines is not enough for a race detector.

alsamylkin:
With DIG loader will be much simpler :)

PR #15, opened 2017-06-01
Simplify configuration loading process

alsamylkin:
The process of loading configs is very convoluted right now: we have interpolated files, non-interpolated files, dynamic providers, functions to look for environment...

We can greatly simplify this process by having a single function that will load all the requested YAML files.


config.go
```diff
@@ -40,287 +38,123 @@ const (
...
-	}
-
-	// Order is important: we want users to be able to override static provider
-	l.RegisterProviders(l.YamlProvider())
-	l.RegisterProviders(providers...)
-
-	return l
+var DefaultLoader = Loader{
+	LookUp: os.LookupEnv,
+	Init:   NewYAMLProviderFromReaderWithExpand(os.LookupEnv, ioutil.NopCloser(strings.NewReader(_defaultInit))),
```
glibsm: This seems bizarre. Why are we using yaml to express loading yaml?
Why can't this be `[]struct{path string; interpolate, optional bool}`?

config.go
```diff
@@ -21,13 +21,11 @@
...
 import (
-	"fmt"
+	"io/ioutil"
 	"os"
-	"path"
-	"path/filepath"
-	"sync"
+	"strings"
 
-	flag "github.com/ogier/pflag"
```
akshayjshah: Why are we using this random fork of Steve Francia's `pflag` library? It's years behind upstream.

Steve's now a core member of the Go team at Google, and he's an upstanding package maintainer. If possible, we should use `spf13/pflag` directly.

alsamylkin: Thanks! We don't use any particular features of that library, all I needed was POSIX compatibility.

config.go
alsamylkin: We are using a config provider, the way it is loaded is through a YAML, but we can use any other provider, it looks cleaner than using a hand crafted types. If someone else wants to use other types, they can use them through config.Populate and control the load flow.

alsamylkin:
@akshayjshah this PR is mostly FYI, to show how we can override config load in private repositories.
One can attach any loading sequences and have lookup functions that are based either on environment variables or just loaded from files.

config.go
alsamylkin: Config shalt configure thyself :)

config.go
```diff
@@ -42,293 +40,98 @@ const (
...
 )
 
-const (
-	_appRoot     = "_ROOT"
-	_environment = "_ENVIRONMENT"
-	_configDir   = "_CONFIG_DIR"
-	_baseFile    = "base.yaml"
-	_secretsFile = "secrets.yaml"
-	_devEnv      = "development"
-)
```
glibsm: this red block seems like a huge win

config.go
```diff
@@ -42,293 +40,98 @@ const (
...
-
-func (l *Loader) getResolver() FileResolver {
-	return NewRelativeResolver(l.Paths()...)
-}
+func LoadDefaults() (Provider, error) {
+	p, err := LoadFromFiles(
+		[]string{"${CONFIG_DIR:config}"},
+		[]FileInfo{
+			{Name: "base.yaml", Interpolate: true},
+			{Name: "${ENVIRONMENT:development}.yaml", Interpolate: true},
```
glibsm: Can we not support interpolation in the actual name of the file?
What actual benefits does this provide, apart from making it easier to test?

alsamylkin: Well, you don't have to do lookups yourself, e.g. environment will be set for you.
I can make it simple though.

akshayjshah:
Sorry, didn't realize that you were waiting on me (if you were). Glad you landed...

alsamylkin:
No worries, I was waiting for input from Glib.

Issue #16, opened 2017-06-05
Defaults don't appear to work with Populate


akshayjshah:
I'm having trouble making the `Value` type's `WithDefault` work with `Populate`. I expect this test to pass, but it doesn't:

```
func TestConfigDefaults(t *testing.T) {
        type cfg struct{ N int }

        var c cfg
        require.NoError(
                t,
                config.NewStaticProvider(nil).Get("metrics").WithDefault(cfg{42}).Populate(&c),
                "Failed to populate config.",
        )
        if c.N != 42 {
                t.Fatalf("Expected to see default N of 42, got %+v.", c)
        }
}
```

Am I using these APIs wrong, or is this a bug?

/cc @blampe (can't get your username to populate in the assignees drop-down)

alsamylkin:
Yeah, WithDefault works only on scalar values right now, so it is very [shallow](https://github.com/uber-go/config/blob/master/value.go#L141). We can do better though and update the root provider to provide a full copy.

akshayjshah:
Fixed!

PR #17, opened 2017-06-05
Make WithDefault to provide a deep copy of the provided default value

alsamylkin:
WithDefaultValue replaces only a default value, which makes it to work only with scalar types, but more complex types will [fail](https://github.com/uber-go/config/issues/16).
We can update the provider saved in value to be a grouped provider of the original provider and a static provider based on the provided value.


glibsm:
Changelog?

alsamylkin:
Config doesn't have a changelog :)

PR #19, opened 2017-06-06
Fix interpolated bools

alsamylkin:
We are using internal `convertValue` that is supposed to work for strings only(to convert from strings to time.Duration/implementation of text.Unmarshaller)
Treating bools as scalars will dispatch decoder to a correct [strconv.ParseBool function].(https://github.com/uber-go/config/blob/master/decoder.go#L190)


value.go
```diff
@@ -195,11 +195,9 @@ func (cv Value) TryAsInt() (int, bool) {
 
 // TryAsBool attempts to return the configuration value as a bool
 func (cv Value) TryAsBool() (bool, bool) {
-	v := cv.Value()
-	if val, err := convertValue(v, reflect.TypeOf(true)); v != nil && err == nil {
-		return val.(bool), true
-	}
-	return false, false
+	var res bool
+	err := newValueProvider(cv.Value()).Get(Root).Populate(&res)
```
glibsm: `vc.Value().Populate`?

static_provider_test.go
```diff
@@ -295,3 +297,17 @@ func TestValue_ChildKeys(t *testing.T) {
...
+	f := func(key string) (string, bool) {
+		if key == "interpolate" {
+			return "true", true
+		}
+		return "", false
+	}
+
+	p := NewYAMLProviderFromReaderWithExpand(f, ioutil.NopCloser(strings.NewReader("val: ${interpolate:false}")))
+	assert.True(t, p.Get("val").AsBool())
+}
```
glibsm: newline

value.go
alsamylkin: I wish :( there is a corner case when a value was created without a provider:
```golang
b := true
v := NewValue(NewStaticProvider(nil), "key", true, true, Bool, nil)
if err := v.Populate(&b); err != nil {
    panic(err)
}
```

glibsm: who is creating values without providers?

alsamylkin: I don't know, some weirdo, who wrote [this test](https://github.com/uber-go/config/blob/master/config_test.go#L182).

glibsm:
Does this fix #18 ?

PR #23, opened 2017-06-08
Use APP_ROOT to resolve configuration path(s)

amckinney:
When using `config.TestConfig` in our application, we were not reading the configuration files from the application root's `config/` directory. This change makes it so that we read from the root if the APP_ROOT variable is set, or the current working directory otherwise.


alsamylkin:
It looks good to me, but I'd suggest to take a look at https://github.com/uber-go/config/pull/15 we are removing loader and most of the global functions with a few simple ones.

amckinney:
Nice! Things are looking good in #15 👍

PR #27, opened 2017-06-20
Check implementation of TextUnmarshaller on all types

alsamylkin:
`Populate` is using `encoding.TextUnmarshaller` interface only on scalar types, 
functions and channels, but any type can implement it, e.g. [zap.AtomicLevel](https://github.com/uber-go/zap/blob/master/level.go#L109)

We are using `Value.String()` to construct the byte array. In case of scalar types
received from a provider, text does not need extra formatting. `TextUnmarshaller`
implementation should be aware of the source of the values for complex types.
For example in `TestUnmarshalTextOnComplexStruct` the actual contents of byte 
array are `map[year:1994 title:Free Willy]`

### DIFF
--- a/decoder.go
+++ b/decoder.go
@@ -456,6 +456,8 @@ func (d *decoder) pointer(name string, value reflect.Value, def string) error {
 	return d.unmarshal(name, value.Elem(), def)
 }
 
+// JSON encoder will fail to serialize maps that don't have strings as keys
+// so we are going to stringify them manually.
 func jsonMap(v interface{}) interface{} {
 	if reflect.TypeOf(v).Kind() == reflect.Map {
 		tmp := make(map[string]interface{})
@@ -470,10 +472,15 @@ func jsonMap(v interface{}) interface{} {
 	return v
 }
 
-func (d *decoder) tryUnmarshallers(key string, value reflect.Value, str string) (bool, error) {
+// tryUnmarshallers checks if the value's type implements either one of standard interfaces in order:
+// 1. `json.Unmarshaler`
+// 2. `encoding.TextUnmarshaler`
+// 3. `yaml.Unmarshaler`
+// and tries it to populate the value.
+func (d *decoder) tryUnmarshalers(key string, value reflect.Value, def string) (bool, error) {
 	switch value.Kind() {
+	// value.IsNil panics if value.Kind() is not equal to one of these constants.
 	case reflect.Chan, reflect.Interface, reflect.Func, reflect.Map, reflect.Ptr:
-		// value.IsNil panics if value.Kind() is not equal to one of these constants:
 		if value.IsNil() {
 			value.Set(reflect.New(value.Type()).Elem())
 		}
@@ -487,34 +494,38 @@ func (d *decoder) tryUnmarshallers(key string, value reflect.Value, str string) 
 	v := d.getGlobalProvider().Get(key)
 	switch t := value.Addr().Interface().(type) {
 	case json.Unmarshaler:
+		// Skip unmarshaling if there is no value.
 		if !v.HasValue() {
 			return true, nil
 		}
 
-		b, err := json.Marshal(jsonMap(v.value))
+		// Marshal the value first.
+		b, err := json.Marshal(jsonMap(v.Value()))
 		if err != nil {
 			return true, errorWithKey(err, key)
 		}
 
+		// Unmarshal corresponding json.
 		return true, errorWithKey(t.UnmarshalJSON(b), key)
 	case encoding.TextUnmarshaler:
+		// check if we need to use custom defaults
 		if v.HasValue() {
-			str = v.String()
+			def = v.String()
 		}
 
-		return true, errorWithKey(t.UnmarshalText([]byte(str)), key)
+		return true, errorWithKey(t.UnmarshalText([]byte(def)), key)
 	case yaml.Unmarshaler:
+		// Skip unmarshaling if there is no value.
 		if !v.HasValue() {
 			return true, nil
 		}
 
-		b, err := yaml.Marshal(v.value)
+		b, err := yaml.Marshal(v.Value())
 		if err != nil {
 			return true, err
 		}
 
-		assign := value.Addr().Interface()
-		return true, errorWithKey(yaml.Unmarshal(b, assign), key)
+		return true, errorWithKey(yaml.Unmarshal(b, value.Addr().Interface()), key)
 	}
 
 	return false, nil
@@ -550,8 +561,8 @@ func (d *decoder) unmarshal(name string, value reflect.Value, def string) error 
 		return errorWithKey(err, name)
 	}
 
-	// Check if a type can be unmarshalled directly.
-	if ok, err := d.tryUnmarshallers(name, value, def); ok {
+	// Check if a type can be unmarshaled directly.
+	if ok, err := d.tryUnmarshalers(name, value, def); ok {
 		return err
 	}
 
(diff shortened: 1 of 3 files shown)


yaml_test.go
```diff
@@ -1054,5 +1072,30 @@ func TestYAMLEnvInterpolationValueConversion(t *testing.T) {
 	v, ok := p.Get("number").TryAsInt()
 	require.True(t, ok)
 	assert.Equal(t, 3, v)
+}
+
+type cartoon struct {
```
glibsm: define the struct in the test itself

decoder.go
```diff
@@ -472,7 +462,9 @@ func (d *decoder) textUnmarshaller(key string, value reflect.Value, str string)
 		return nil
 	}
 
-	if value.IsNil() {
+	kind := value.Kind()
+
+	if kind >= reflect.Array && kind <= reflect.Slice && value.IsNil() {
```
glibsm: i don't know what types are in between the Array and Slice, is there a way to be more explicit here?

yaml_test.go
alsamylkin: I need this struct to implement the `encoding.TextUnmarshaller` interface. I don't know how can we define a method inside a function..

decoder.go
alsamylkin: This was the most concise way I could make this expression: https://golang.org/src/reflect/type.go#L244 . Another option would be to have 6 comparisons.. This is needed only because of the check for nil will panic if it is called on other than these types.

glibsm: I mean at least a comment is required then, since it's completely non-obvious what's going on without jumping into the source of the kind definition

yaml_test.go
glibsm: makes sense

alsamylkin:
I think I can go further and check all possible unmarshallers(json and yaml) before trying to populate separate types.

decoder.go
akshayjshah: A switch might be a nice, readable alternative.

decoder.go
```diff
@@ -454,39 +456,79 @@ func (d *decoder) pointer(name string, value reflect.Value, def string) error {
...
-	return d.textUnmarshaller(key, value, def)
-}
+// JSON encoder will fail to serialize maps that don't have strings as keys
+// so we are going to stringify them manually.
+func jsonMap(v interface{}) interface{} {
+	if reflect.TypeOf(v).Kind() == reflect.Map {
+		tmp := make(map[string]interface{})
+		rv := reflect.ValueOf(v)
+		for _, key := range rv.MapKeys() {
+			tmp[fmt.Sprint(key.Interface())] = jsonMap(rv.MapIndex(key).Interface())
```
glibsm: str, ok := `key.Interface().(string)`?

decoder.go
```diff
@@ -454,39 +456,79 @@ func (d *decoder) pointer(name string, value reflect.Value, def string) error {
...
-
-// Sets value to a function type.
-func (d *decoder) function(key string, value reflect.Value, def string) error {
-	return d.textUnmarshaller(key, value, def)
-}
+// JSON encoder will fail to serialize maps that don't have strings as keys
+// so we are going to stringify them manually.
+func jsonMap(v interface{}) interface{} {
+	if reflect.TypeOf(v).Kind() == reflect.Map {
+		tmp := make(map[string]interface{})
```
glibsm: pre-allocate length?

decoder.go
```diff
@@ -454,39 +456,79 @@ func (d *decoder) pointer(name string, value reflect.Value, def string) error {
...
+		if value.IsNil() {
+			value.Set(reflect.New(value.Type()).Elem())
+		}
 
-	// Value has to have a pointer receiver to be able to modify itself with TextUnmarshaller
-	if !value.CanAddr() {
-		return errorWithKey(errors.New("can't use TextUnmarshaller because value is not addressable"), key)
+	//TODO(alsam) Command-line provider implementation is made via maps, so we'll fail to merge slices and maps
+	// TestCommandLineProvider_NestedValues will fail in that case.
+	case reflect.Slice:
```
glibsm: I'm confused about this. Why does the encoder need to know about specifics of a provider, and when/how do we plan to fix?

value.go
```diff
@@ -286,13 +285,9 @@ func convertValue(value interface{}, targetType reflect.Type) (interface{}, erro
 	switch v := value.(type) {
```
glibsm: comment for the `convertValue` function is now out of date, right?

decoder.go
```diff
@@ -454,39 +456,79 @@ func (d *decoder) pointer(name string, value reflect.Value, def string) error {
...
+	v := d.getGlobalProvider().Get(key)
 	switch t := value.Addr().Interface().(type) {
+	case json.Unmarshaler:
+		// Skip unmarshaling if there is no value.
+		if !v.HasValue() {
+			return true, nil
+		}
+
+		// Marshal the value first.
+		b, err := json.Marshal(jsonMap(v.Value()))
```
glibsm: What is the actual thing inside `v.Value()` right now - is it bytes? is it a string?
Marhsaling first to ask someone else to Unmarshal right away looks really odd to me, but I don't understand all the internals

decoder.go
alsamylkin: Sure :)

decoder.go
alsamylkin: Unfortunately we can't do this: a key could be of any type, e.g. integer. The reason we have to stringify map keys is because JSON marshaller supports only maps with string keys.
Another option would be to return an error, but I want to do our best first and try to populate with a new map instead of simply returning an error.

decoder.go
alsamylkin: Command line provider implements slices as maps internally.
When user asks to populate e.g. `roles` slice it will return a map and group provider will fail to merge it with an array. Decoder's slice method skips request for `roles` key and starts populating elements from the very beginning: `roles.0`, `roles.1` ...

Let me create an issue for this TODO.

decoder.go
alsamylkin: it is an `interface{}`. We don't really know what a provider returns us: it can be a string, or it can be a `map[interface{}]interface{}`(in case of a yaml provider for structs).
The reason is we first doing a marshal first and unmarshal after is that we want to serialize this custom type first and then decode for example to a struct, basically doing a poor man's `mapstructure` implementation.

value.go
alsamylkin: Why? It doesn't mention unmarshaller..

decoder.go
alsamylkin: https://github.com/uber-go/config/issues/29

decoder.go
```diff
@@ -454,39 +456,79 @@ func (d *decoder) pointer(name string, value reflect.Value, def string) error {
...
+		return true, errorWithKey(t.UnmarshalText([]byte(def)), key)
+	case yaml.Unmarshaler:
+		// Skip unmarshaling if there is no value.
+		if !v.HasValue() {
+			return true, nil
+		}
+
+		b, err := yaml.Marshal(v.Value())
+		if err != nil {
+			return true, err
```
glibsm: line not covered

decoder.go
```diff
@@ -519,8 +561,14 @@ func (d *decoder) unmarshal(name string, value reflect.Value, def string) error
...
+	// Check if a type can be unmarshaled directly.
+	if ok, err := d.tryUnmarshalers(name, value, def); ok {
+		return err
+	}
+
+	// Try to unmarshal each type separately.
 	switch value.Kind() {
-	case reflect.Invalid:
+	case reflect.Invalid, reflect.Chan, reflect.Func:
 		return fmt.Errorf("invalid value type for key %s", name)
```
glibsm: also come uncovered lines

PR #33, opened 2017-06-27
Try methods should return false for missing values

alsamylkin:
Title is self explanatory..

### DIFF
--- a/static_provider_test.go
+++ b/static_provider_test.go
@@ -343,3 +343,29 @@ func TestConfigDefaultsAreOverriddenByHigherPriorityProviders(t *testing.T) {
 	)
 	assert.Equal(t, book{Title: "The Financier", Author: "Dreiser", Year: 1925}, novel)
 }
+
+func TestMissingValuesForTryMethods(t *testing.T) {
+	t.Parallel()
+
+	assert := assert.New(t)
+	p := NewStaticProvider(nil)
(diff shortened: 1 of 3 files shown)


static_provider_test.go
```diff
@@ -343,3 +343,29 @@ func TestConfigDefaultsAreOverriddenByHigherPriorityProviders(t *testing.T) {
...
 	assert.Equal(t, book{Title: "The Financier", Author: "Dreiser", Year: 1925}, novel)
 }
+
+func TestMissingValuesForTryMethods(t *testing.T) {
+	t.Parallel()
+
+	assert := assert.New(t)
+	p := NewStaticProvider(nil)
+	t.Run("int", func(t *testing.T) {
+		_, ok := p.Get("missing").TryAsInt()
```
glibsm: `v := p.Get("missing")` can happen outside of these tests

static_provider_test.go
```diff
@@ -343,3 +343,29 @@ func TestConfigDefaultsAreOverriddenByHigherPriorityProviders(t *testing.T) {
...
+		assert.False(ok)
+	})
+
+	t.Run("float", func(t *testing.T) {
```
glibsm: table-driven as opposed to sub-test would be much easier to write and maintain I think, but it's up to you

PR #37, opened 2017-07-14
Skip inappropriate types instead of reporting errors on them

alsamylkin:
Sometimes configuration structs can contain public fields that are channels/functions that don't implement TextUnmarshaller/UnmarshalYAML interfaces. They are are supposed to be field manually/by another function and errors break this flow. We should silently skip them instead, an example is [json.Unmarshal](https://golang.org/pkg/encoding/json/#Unmarshal)

### DIFF
--- a/decoder.go
+++ b/decoder.go
@@ -623,6 +623,10 @@ func (d *decoder) unmarshal(name string, value reflect.Value, def string) error 
 	switch value.Kind() {
 	case reflect.Invalid:
 		return fmt.Errorf("invalid value type for key %s", name)
+
+	// Permissive decoding: skip Chan/Func fields, instead of returning an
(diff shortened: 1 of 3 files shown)


decoder.go
```diff
@@ -621,8 +621,10 @@ func (d *decoder) unmarshal(name string, value reflect.Value, def string) error
 
 	// Try to unmarshal each type separately.
 	switch value.Kind() {
-	case reflect.Invalid, reflect.Chan, reflect.Func:
+	case reflect.Invalid:
 		return fmt.Errorf("invalid value type for key %s", name)
+	case reflect.Chan, reflect.Func:
+		return nil
```
glibsm: lets leave a comment here explaining the reasoning

PR #41, opened 2017-07-18
Use semantic version paths for yaml and validator packages

alsamylkin:
Most of the projects are using semantic version paths(e.g. yarpc, dosa, tally),
so it is better to be in sync with them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
   `reflect.Kind`.
 - Skip populating function and value types instead of reporting errors.
 - **[Breaking]** `Value.Timestamp` is private, use Value.LastUpdated instead.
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
   `reflect.Kind`.
 - Skip populating function and value types instead of reporting errors.
 - **[Breaking]** `Value.Timestamp` is private, use Value.LastUpdated instead.
+- **[Breaking]** Use semantic version paths for yaml and validator packages.
```
glibsm: is it actually breaking?

alsamylkin: Yes, you need to go get a gopkg.in package, not a github/go-yaml.

PR #42, opened 2017-07-19
Let user to skip loading command line provider via  parameter.

alsamylkin:
Right now there is no easy way of skipping command line provider which can conflict with another command line libraries. Instead of requiring user to set up all the flags in the `pflag.CommandLine` variable we'll read `commandline` value from previously loaded providers(yaml files).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
   `reflect.Kind`.
 - Skip populating function and value types instead of reporting errors.
 - **[Breaking]** `Value.Timestamp` is private, use Value.LastUpdated instead.
+- Let user to skip loading command line provider via `commandline` parameter.
 
 ## v1.0.0-rc1 (26 Jun 2017)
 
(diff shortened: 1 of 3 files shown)


glibsm:
the documentation is confusing as far as *where* to specify the command line skip

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - Skip populating function and value types instead of reporting errors.
 - **[Breaking]** `Value.Timestamp` is private, use Value.LastUpdated instead.
 - **[Breaking]** Use semantic version paths for yaml and validator packages.
+- Let user to skip loading command line provider via `commandline` parameter.
```
glibsm: `commandLine`? Camel it, since it's two words

config.go
```diff
@@ -43,7 +43,10 @@ type FileInfo struct {
 //  ./config/${ENVIRONMENT}.yaml
 //  ./config/secrets.yaml
 // The provider is also amended with a command line provider that reads
-// a roles flag.
+// a roles flag. It can be skipped by seting commandline value to false
+// at the root level of config file:
```
glibsm: we don't refer to the `root level of config file` anywhere else. Just call out `base.yaml`

config.go
```diff
@@ -43,7 +43,10 @@ type FileInfo struct {
...
+//
+//  commandline: false
```
glibsm: code samples are indented

Issue #24, opened 2017-06-12
Errors over panics


glibsm:
Config code panics a lot when something doesn't go it's way: can't open yaml files, can't marshal, etc.

Before start the release cycle I'd like for all the providers to return `(Provider, error)`.

For example, one of the must uses providers: `func NewYAMLProviderFromFiles(files ...string) Provider` has multiple ways it can panic.

I get that configuration loading is a pretty crucial process and more often than not, users will chose to panic right away. However, let's not make that choice for them and at least give them the option to do something else.

We're a library, and like dig, we should try very very hard to reserve panics for unexpected conditions, rather than trying to use it as an exception.

alsamylkin:
via https://github.com/uber-go/config/pull/43

PR #43, opened 2017-07-20
No panics in config constructors

alsamylkin:
Related to https://github.com/uber-go/config/issues/24

Instead of panicking on errors, we should simply return them
to a caller end let them decide what to do with the errors.

The change is mostly straight forward, but there is one snowflake:
expand variables. The replace function doesn't return an error,
so we panic internally and use recover to report errors.

### DIFF
--- a/command_line_provider.go
+++ b/command_line_provider.go
@@ -56,9 +56,9 @@ type commandLineProvider struct {
 // arguments as config values. In order to address nested elements one
 // can use dots in flag names which are considered separators.
 // One can use StringSlice type to work with a list of comma separated strings.
-func NewCommandLineProvider(flags *flag.FlagSet, args []string) Provider {
+func NewCommandLineProvider(flags *flag.FlagSet, args []string) (Provider, error) {
 	if err := flags.Parse(args); err != nil {
-		panic(err)
+		return nil, err
 	}
 
 	m := make(map[string]interface{})
@@ -77,7 +77,12 @@ func NewCommandLineProvider(flags *flag.FlagSet, args []string) Provider {
 		m[f.Name] = f.Value.String()
 	})
 
-	return commandLineProvider{Provider: NewStaticProvider(m)}
+	p, err := NewStaticProvider(m)
+	if err != nil {
+		return nil, err
+	}
+
+	return commandLineProvider{Provider: p}, nil
 }
 
 // Name implements the Provider interface.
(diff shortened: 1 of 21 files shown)


yaml.go
```diff
@@ -285,18 +320,29 @@ func (n *yamlNode) Children() []*yamlNode {
...
 
-func (n *yamlNode) applyOnAllNodes(expand func(string) string) {
-	if n == nil {
-		return
-	}
+func (n *yamlNode) applyOnAllNodes(expand func(string) string) (err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			if _, ok := r.(runtime.Error); ok {
+				fmt.Println(r)
```
glibsm: no need to print before panic here

command_line_provider.go
```diff
@@ -77,7 +77,12 @@ func NewCommandLineProvider(flags *flag.FlagSet, args []string) Provider {
 		m[f.Name] = f.Value.String()
 	})
 
-	return commandLineProvider{Provider: NewStaticProvider(m)}
+	p, err := NewStaticProvider(m)
+	if err != nil {
+		return nil, err
```
glibsm: not covered

static_provider.go
```diff
@@ -35,33 +35,48 @@ type staticProvider struct {
...
-		Provider: NewYAMLProviderFromReader(toReadCloser(data)),
+func NewStaticProvider(data interface{}) (Provider, error) {
+	c, err := toReadCloser(data)
+	if err != nil {
+		return nil, err
 	}
+
+	p, err := NewYAMLProviderFromReader(c)
+	if err != nil {
+		return nil, err
```
glibsm: also not covered

yaml.go
```diff
@@ -285,18 +320,29 @@ func (n *yamlNode) Children() []*yamlNode {
...
 }
 
-func (n *yamlNode) applyOnAllNodes(expand func(string) string) {
-	if n == nil {
-		return
-	}
+func (n *yamlNode) applyOnAllNodes(expand func(string) string) (err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			if _, ok := r.(runtime.Error); ok {
```
glibsm: do we want to explicitly filter out the runtime errors in this case?

alsamylkin: yes, runtime.Error is returned for nil pointer exceptions/index out of bounds. Config internal errors are plain builtin errors.

yaml.go
alsamylkin: oops..

command_line_provider.go
alsamylkin: I can cover this line, but it'll not provide much value, simply a public function returning an error via another public function.

static_provider.go
alsamylkin: This is almost impossible to cover and doesn't provide much value.

alsamylkin:
image

PR #44, opened 2017-07-25
Simplify the Initial API

amckinney:
**The goal of this pull request is to simplify the configuration solution so that it focuses on a smaller initial API.**

**Suggested Changes:**

**(1) Remove the Command Line Provider**

- The command line provider is not be used in the vast majority of cases.

- Removes the StringSlice type.

- We can bring this functionality back if users request it, but the library will be easier to maintain without it.

**(2) Un-export the Cache Provider**

- The cached provider is great for improving the performance of our YAML providers, but we don’t expect users to interact with it directly.

- Users should be accessing a smaller set of exposed YAML provider solutions.

**(3) Remove the config.Value.ChildKeys Method**

- The **ChildKeys** method is not used in any of the internal functions. In our customers’ experience, we have never needed to access a value’s child keys. 

**(4) Remove the config.Value.timestamp attribute and config.Value.LastUpdated Method**

- If we make this configuration library more focused on serving static YAML files, there is no need for maintaining timestamps in our values. 

- All internal references to **config.NewValue** passed nil for the value of the given timestamp, so removing the attribute was seemingly harmless.

- We will not be reading updates so we can also remove the **LastUpdated** method. 

**(5) Update Godocs for Consistency**

- There are minor discrepancies between godocs on a variety of different methods. E.g. Comments not ending with periods.


breerly:
@alsamylkin we're going to be living with this package for a *long* time.

Can reduce the library as much as possible, and then extend it internally for the usecases you mentioned? We did the same thing with yarpc, and it's worked out really well.

I want to make sure that our foundational building blocks are as small as possible. I'm 100% in favor of what Alex M is trying to do in spirit here - we can always add to this lib later, but we'll never be able to take away.

alsamylkin:
@breerly I am all up for reducing the surface. If it is a complex interface - definitely, if it is a simple function(that command line provider and cache provider really are from API perspective), then it looks a bit too much. Just look what is removed: only these files, there are no extra dependencies on these providers. I also don't see much value keeping them internal, because they are built from open components and it is so much better to keep everything public so we build easily customizable APIs.

We discussed the ChildKeys method with @glibsm before and decided to keep it, because addressing values in providers is still an open question and it actually simplifies looking for keys on the provider side, not on the consumer. Current dot separator is ok, but is far from perfect.

blampe:
"When in doubt, leave it out." We can always add things but we can never remove them. Let's stick to the bare minimum feature set we need to get things stable and working end-to-end.

glibsm:
I have no objections to the _spirit_ of this PR - less is infinitely more. However, I would have liked to see _an issue on the subject created first_ for the discussion purposes (which is what is happening on this PR now).

The main question I have is what are we going to do to the users of both the ChildKeys (catalyst) and command line providers (which is an internally documented and used feature of config)?

We're not operating under the assumption that nobody is using this stuff, we're trying to remove wheels from a moving train, albeit without _too_ many passengers. 

In either case, I would like to take this offline. 30m discussion should save us all a lot of typing.

breerly:
Hard to argue - I assumed that y'all jammed before this PR was open. If that didn't happen, that's our fault, we always want to have inclusive alignment-building jam sheshs leading up to these type of PRs.

Let's get together and do it right.

breerly:
PS - since I'm in AMS till next week, ya'll should jam without me!

Issue #32, opened 2017-06-27
Drop the conversion helpers from the the Value struct


glibsm:
As we're shooting towards the v1 release, the number of public methods on the Value type has been called into question.

## Value methods
```
type Value
    // Keep:
    func (cv Value) ChildKeys() []string
    func (cv Value) HasValue() bool
    func (cv Value) IsDefault() bool
    func (cv Value) LastUpdated() time.Time
    func (cv Value) Populate(target interface{}) error
    func (cv Value) Get(key string) Value
    func (cv Value) Source() string
    func (cv Value) Value() interface{}
    func (cv Value) WithDefault(value interface{}) Value

    // Decision zone ("helpers"):
    func (cv Value) AsBool() bool
    func (cv Value) AsFloat() float64
    func (cv Value) AsInt() int
    func (cv Value) AsString() string
    func (cv Value) String() string
    func (cv Value) TryAsBool() (bool, bool)
    func (cv Value) TryAsFloat() (float64, bool)
    func (cv Value) TryAsInt() (int, bool)
    func (cv Value) TryAsString() (string, bool)
```
Godoc: https://godoc.org/go.uber.org/config#Value

## Reasoning

Most of these helpers look very similar. They rely heavily on the `Populate()` functionality and try to mimic what the users would otherwise do should these methods not be available.

Here are a subset of two:

```go
// TryAsBool attempts to return the configuration value as a bool
func (cv Value) TryAsBool() (bool, bool) {
	var res bool
	err := newValueProvider(cv.Value()).Get(Root).Populate(&res)
	return res, err == nil
}

// TryAsFloat attempts to return the configuration value as a float
func (cv Value) TryAsFloat() (float64, bool) {
	var res float64
	err := newValueProvider(cv.Value()).Get(Root).Populate(&res)
	return res, err == nil
}
```

### A

Do nothing. As ever this is the first option.

The main problem for me is that we have a strange subset of types that we consider "worthy" of the own method. What about `float32`, what about `int16`? What about a `[]int`? 

If we go down this road we'd need some sort of a heuristic to determine what warrants a conversion method.

### B

Move these methods into functions in another package. `func (cv Value) TryAsInt()` can become `func TryAsInt(v Value)`.

For the sake of discussion lets name it `package conv`.

This is a very attractive option, as we can have a more comprehensive coverage of available functions and types without diluting the API of `Value struct`.

[...]

PR #45, opened 2017-07-26
Remove casts

alsamylkin:
Related to this issue:
https://github.com/uber-go/config/issues/32

We can use different library, e.g. https://github.com/spf13/cast to do casting,
because supports more types than we do now.


glibsm:
Can we write some integration tests in another test package with one said library, a-la zap benchmarks?

Would not only be good to illustrate the usage (without incurring a dependency for users), but also to make sure we remain compatible and can tweak our API to suit that use-case better.

I believe right now it involves something like calling `Value.Value()` =/

command_line_provider_test.go
```diff
@@ -71,7 +71,7 @@ func TestCommandLineProvider_Conversion(t *testing.T) {
 
 	v := c.Get("dozen")
 	require.True(t, v.HasValue())
-	assert.Equal(t, 13, v.AsInt())
+	assert.Equal(t, "13", v.String())
```
abhinav: What about

    assert.Equal(t, 13, v.Value())

alsamylkin: It is going to be a string, not a integer. Same as environment variables:
```golang
assert.Equal(t, "13", v.Value())
```

abhinav: Ah, okay.

PR #46, opened 2017-07-28
More examples to Value.Populate and NewProviderGroup functions

alsamylkin:
Add more examples to `Value.Populate` and `NewProviderGroup` functions.


amckinney:
Curious, why don't we use `assert` or `require` rather than `fmt` in these tests?

alsamylkin:
@amckinney Examples are tested differently than regular tests: https://blog.golang.org/examples

PR #47, opened 2017-07-28
Remove Load* function

alsamylkin:
We don't really need Load* functions, because users can use constructors directly to load files.
File name to load config, that is based on the environment variable is also too company specific.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@
 - **[Breaking]** Try and As conversion helpers are removed in favor of using
   other cast libraries.
 - **[Breaking]** Removed `Value.IsDefault` method.
-- **[Breaking]** Removed Load* functions because they are too Uber specific.
+- **[Breaking]** Removed Load* functions.
 
 ## v1.0.0-rc1 (26 Jun 2017)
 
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 - **[Breaking]** `Value.WithDefault` returns an error when a default can't be used.
 - **[Breaking]** Try and As conversion helpers are removed in favor of using
   other cast libraries.
+- **[Breaking]** Removed Load* functions because they are too Uber specific.
```
abhinav: changelog doesn't ned justification. Just say,

    Removed `Load*` functions.

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@
 - **[Breaking]** Try and As conversion helpers are removed in favor of using
   other cast libraries.
 - **[Breaking]** Removed `Value.IsDefault` method.
+- **[Breaking]** Removed Load* functions.
+- **[Breaking]** Removed NewYAMLProviderFromReader* functions.
```
amckinney: Consider changing `Removed` to `Unexport` for the `NewYAMLProviderFromReader*` functions.

PR #48, opened 2017-07-28
Merge values for provider group in the constructor

alsamylkin:
ProviderGroup can and should merge values on construction, instead of on value retrieval.
To do that we need need apply expand to all nested YAML nodes, instead of only top level ones.


glibsm:
What about `p.Get("foo").Value()` now that we're driving people to the casting libraries? Sounds like they will just get the default value, rather than exposing the error

alsamylkin:
Yeah, there are some other corner cases while populating values inside,  e.g. there is a merge error for populating a first element in the `[]interface{}` slice. Need to think more about it.

provider_group.go
```diff
@@ -21,56 +21,44 @@
...
 // A and B and values from B will overwrite values of A. e.g.
 //   A:                B:                 merge(A, B):
 //     keep:A            new:B              keep:A
 //     update:fromA      update:fromB       update:fromB
 //                                          new:B
 //
-// * if A is a map and B is not, this function will panic,
-// e.g. key:value and -slice
+// If A is a map and B is not, this function will return a Value with
+// an error inside.
```
glibsm: Is there a case when there is a value with an error inside that will then propagate it to `Populate()` and `Value()` method? Or this comment now obsolete

alsamylkin: It can be an error value from the provider, not a merge error. We construct only correct providers now, no runtime surprises.

breerly:
@alsamylkin Is this WIP or ready for review?

alsamylkin:
It is ready for review.

yaml.go
```diff
@@ -82,7 +82,8 @@ func newYAMLProviderCore(files ...io.ReadCloser) (*yamlConfigProvider, error) {
 //     update:fromA      update:fromB       update:fromB
 //                                          new:B
 //
-// * if A is a map and B is not, this function will panic, e.g. key:value and -slice
+// * if A is a map and B is not, this function will return an error,
```
amckinney: For consistency, consider updating `* if` to `If` same as in `* in` below.

value.go
```diff
@@ -71,7 +71,11 @@ func (cv Value) WithDefault(value interface{}) (Value, error) {
 		return cv, err
 	}
 
-	g := NewProviderGroup("withDefault", p, cv.provider)
+	g, err := NewProviderGroup("withDefault", p, cv.provider)
+  if err != nil {
```
glibsm: formatting is odd here

alsamylkin: This is an only merge tool.

PR #49, opened 2017-07-28
Remove dead wood

alsamylkin:
Looks like a https://github.com/uber-go/config/pull/44 is stuck in a deadlock.
Lets remove at least parts that we agree upon.


alsamylkin:
Closed in favor of https://github.com/uber-go/config/pull/44

Issue #55, opened 2017-08-04
[Regression] Populate attempts to unmarshal keys that aren't in the provider


bayesianmind:
In previous versions of config values where HasValue == false were never deserialized (theory, haven't confirmed that). In 1.0.0 every field is getting deserialized, even if it has no corresponding value in the yaml.

This breaks deserializing into a struct like xtchannel.Configuration which contains `UnmarshalText` methods that don't accept default values as valid.

alsamylkin:
Via https://github.com/uber-go/config/pull/56

PR #56, opened 2017-08-04
Fix unmarshal text on missing value.

alsamylkin:
We invoke `TextUnmarshaler.UnmarshalText` even if a value is missing from a config, which will lead to errors like https://github.com/uber-go/config/issues/55. We should keep a zero initialized value instead of unmarshalling an empty string.

### DIFF
--- a/decoder.go
+++ b/decoder.go
@@ -529,7 +529,8 @@ func jsonMap(v interface{}) interface{} {
 	return v
 }
 
-func skip(value *Value, def string) bool {
+// Skip unmarshaling if there is no value in provider and default is empty.
+func shouldSkip(value *Value, def string) bool {
 	return !value.HasValue() && def == ""
 }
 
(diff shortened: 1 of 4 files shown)


decoder.go
```diff
@@ -561,18 +570,26 @@ func (d *decoder) tryUnmarshalers(key string, value reflect.Value, def string) (
 		// Unmarshal corresponding json.
 		return true, errorWithKey(t.UnmarshalJSON(b), key)
 	case encoding.TextUnmarshaler:
+		if skip(&v, def) {
```
glibsm: why not do all these outside the switch?

decoder.go
```diff
@@ -548,10 +552,15 @@ func (d *decoder) tryUnmarshalers(key string, value reflect.Value, def string) (
 	switch t := value.Addr().Interface().(type) {
 	case json.Unmarshaler:
 		// Skip unmarshaling if there is no value.
-		if !v.HasValue() {
+		if skip(&v, def) {
 			return true, nil
 		}
 
+		// check if we need to use custom defaults
```
glibsm: this comment means nothing to me

decoder.go
```diff
@@ -529,6 +529,10 @@ func jsonMap(v interface{}) interface{} {
 	return v
 }
 
+func skip(value *Value, def string) bool {
```
glibsm: `shouldSkip`, skip implies action

decoder.go
alsamylkin: if we don't have a value from the config provider, we should use a default.

decoder.go
alsamylkin: because it is a type switch, and we want to return false in case if a type is not one of the unmarshalers

PR #58, opened 2017-08-10
Perform variable expansion using text.Transformer

drafnel:
Currently NewYAMLProviderFromFiles() and NewYAMLProviderWithExpand() do not produce the same result when given the same input, even when the input contains only simple values with no expansion variables.  The expansion feature has a side-effect which converts all fields into strings.

This can be demonstrated with:

    package main

    import "fmt"
    import "os"

    import "go.uber.org/config"

    func main() {
        if len(os.Args) != 3 {
            fmt.Fprintf(os.Stderr, "Usage: %s config.yaml field\n",
                os.Args[0])
            os.Exit(1)
        }

        filename := os.Args[1]
        field := os.Args[2]

        cfg, _ := config.NewYAMLProviderFromFiles(filename)
        val := cfg.Get(field)
        fmt.Printf("%s: %v (%T)\n", field, val.Value(), val.Value())

        cfg, _ = config.NewYAMLProviderWithExpand(os.LookupEnv, filename)
        val = cfg.Get(field)
        fmt.Printf("%s: %v (%T)\n", field, val.Value(), val.Value())
    }

    $ cat test.yaml
    foo:
        bar:
            baz: 13`

    $ ./get-config test.yaml foo.bar.baz
    foo.bar.baz: 13 (int)
    foo.bar.baz: 13 (string)

This series introduces a text.Transformer to perform the variable expansion via an io.ReadCloser so it can be transparently used by newYAMLProviderCore().  It also introduces the escape sequence '$$' which will be replaced by a literal '$'.

fyi @akshayjshah @breerly 

-Brandon

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,10 @@
 
 ## v1.1.0 (not released yet)
 
-- No changes yet.
+- Expand functions transform a special sequence $$ to literal $.
+- The underlying objects encapsulated by config.Value types will now
+  have the types determined by the YAML unmarshaller regardless of
+  whether expansion was performed or not.
 
 ## v1.0.2 (2017-08-17)
 
--- a/expand.go
+++ b/expand.go
@@ -1,0 +1,197 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package config
+
+import (
+	"bytes"
+
+	"golang.org/x/text/transform"
+)
+
+// expandTransformer implements transform.Transformer
+type expandTransformer struct {
+	transform.NopResetter
+	expand func(string) (string, error)
+}
+
+// First char of shell variable may be [a-zA-Z_]
+func isShellNameFirstChar(c byte) bool {
+	return c == '_' ||
+		(c >= 'a' && c <= 'z') ||
+		(c >= 'A' && c <= 'Z')
+}
+
+// Char's after the first of shell variable may be [a-zA-Z0-9_]
+func isShellNameChar(c byte) bool {
+	return c == '_' ||
+		(c >= 'a' && c <= 'z') ||
+		(c >= 'A' && c <= 'Z') ||
+		(c >= '0' && c <= '9')
+}
+
+// bytesIndexCFunc returns the index of the byte for which the
+// complement of the supplied function is true
+func bytesIndexCFunc(buf []byte, f func(b byte) bool) int {
+	for i, b := range buf {
+		if !f(b) {
+			return i
+		}
+	}
+	return -1
+}
+
+// Transform expands shell-like sequences like $foo and ${foo} using
+// the configured expand function.  The sequence '$$' is replaced with
+// a literal '$'.
+func (e *expandTransformer) Transform(dst, src []byte, atEOF bool) (int, int, error) {
+	var srcPos int
+	var dstPos int
+
+	for srcPos < len(src) {
+
+		if dstPos == len(dst) {
+			return dstPos, srcPos, transform.ErrShortDst
+		}
+
+		end := bytes.IndexByte(src[srcPos:], '$')
+
+		if end == -1 {
+			// src does not contain '$', copy into dst
+			cnt := copy(dst[dstPos:], src[srcPos:])
+			srcPos += cnt
+			dstPos += cnt
+			continue
+		} else if end > 0 {
+			// copy chars preceding '$' from src to dst
+			cnt := copy(dst[dstPos:], src[srcPos:srcPos+end])
+			srcPos += cnt
+			dstPos += cnt
+
+			if dstPos == len(dst) {
+				return dstPos, srcPos, transform.ErrShortDst
+			}
+		}
+
+		// src[srcPos] now points to '$', dstPos < len(dst)
+
+		// If we're at the end of src, but we found a starting
+		// token, return ErrShortSrc, unless we're also at EOF,
+		// in which case just copy it dst.
+		if srcPos+1 == len(src) {
+			if atEOF {
+				dst[dstPos] = src[srcPos]
+				srcPos++
+				dstPos++
+				continue
+			}
+			return dstPos, srcPos, transform.ErrShortSrc
+		}
+
+		// At this point we know that src[srcPos+1] is populated.
+
+		// If this token sequence represents the special '$$'
+		// sequence, emit a '$' into dst.
+		if src[srcPos+1] == '$' {
+			dst[dstPos] = src[srcPos]
+			srcPos += 2
+			dstPos++
+			continue
+		}
+
+		var token []byte
+		var tokenEnd int
+
+		// Start of bracketed token ${foo}
+		if src[srcPos+1] == '{' {
+			end := bytes.IndexByte(src[srcPos+2:], '}')
+			if end == -1 {
+				if atEOF {
+					// No closing bracket and we're at
+					// EOF, so it's not a valid bracket
+					// expression.
+					if len(dst[dstPos:]) <
+						len(src[srcPos:]) {
+						return dstPos, srcPos,
+							transform.ErrShortDst
+					}
+
+					cnt := copy(dst[dstPos:], src[srcPos:])
+					srcPos += cnt
+					dstPos += cnt
+					continue
+				}
+
+				// Otherwise, we need more bytes in src
+				return dstPos, srcPos, transform.ErrShortSrc
+			}
+
+			// Set tokenEnd so it points to the byte
+			// immediately after the closing '}'
+			tokenEnd = end + srcPos + 3
+
+			token = src[srcPos+2 : tokenEnd-1]
+		} else { // Else start of non-bracketed token $foo
+			if !isShellNameFirstChar(src[srcPos+1]) {
+				// If it doesn't conform to the naming
+				// rules for shell variables, do not
+				// try to expand, just copy to dst.
+				dst[dstPos] = src[srcPos]
+				srcPos++
+				dstPos++
+				continue
+			}
+
+			end := bytesIndexCFunc(src[srcPos+2:], isShellNameChar)
+
+			if end == -1 {
+				// Reached the end of src without finding
+				// end of shell variable
+				if !atEOF {
+					// We need more bytes in src
+					return dstPos, srcPos,
+						transform.ErrShortSrc
+				}
+				tokenEnd = len(src)
+			} else {
+				// Set tokenEnd so it points to the byte
+				// immediately after the token
+				tokenEnd = end + srcPos + 2
+			}
+
+			token = src[srcPos+1 : tokenEnd]
+		}
+
+		replacement, err := e.expand(string(token))
+		if err != nil {
+			return dstPos, srcPos, err
+		}
+
+		if len(dst[dstPos:]) < len(replacement) {
+			return dstPos, srcPos, transform.ErrShortDst
+		}
+
+		cnt := copy(dst[dstPos:], replacement)
+		srcPos = tokenEnd
+		dstPos += cnt
+	}
+
+	return dstPos, srcPos, nil
+}
--- a/expand_test.go
+++ b/expand_test.go
@@ -1,0 +1,295 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package config
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/text/transform"
+)
+
+// Size of buffer used by the transform package.
+const transformBufSize = 4096
+
+const orig = `This is a $t3sT$. $$ This is a $$test.
+	This is not a valid $0ne.  But this one $i5_@_valid-one.
+	$$$$$$$
+${parti`
+
+const expected = `This is a test$. $ This is a $test.
+	This is not a valid $0ne.  But this one is @_valid-one.
+	$$$$
+${parti`
+
+const ends_in_dollar = `There is a dollar at the end$`
+const ends_in_ddollar = `There is a dollar at the end$$`
+const many_dollars_orig = `$$$$$$$`
+const many_dollars_expect = `$$$$`
+const ends_in_var = `There is a test at the end: $t3sT`
+const ends_in_var_expect = `There is a test at the end: test`
+
+type oneByteReader struct {
+	r io.Reader
+}
+
+func (e *oneByteReader) Read(buf []byte) (n int, err error) {
+	var b [1]byte
+
+	if len(buf) > 0 {
+		n, err = e.r.Read(b[:])
+		buf[0] = b[0]
+	}
+
+	return
+}
+
+type bufReader struct {
+	buf    []byte
+	offset int
+}
+
+// Similar to a bytes.Reader except this Reader returns EOF in the same
+// Read that reads the end of the buffer.
+func (e *bufReader) Read(buf []byte) (int, error) {
+	var err error
+	n := copy(buf, e.buf[e.offset:])
+	e.offset += n
+	if e.offset == len(e.buf) {
+		err = io.EOF
+	}
+	return n, err
+}
+
+func TestExpander(t *testing.T) {
+	t.Parallel()
+
+	r := bytes.NewReader([]byte(orig))
+
+	expand_func := func(s string) (string, error) {
+		switch s {
+		case "t3sT":
+			return "test", nil
+		case "i5_":
+			return "is ", nil
+		}
+
+		return "NOMATCH", errors.New("No Match")
+	}
+
+	// Parse whole string
+	tr := transform.NewReader(r, &expandTransformer{expand: expand_func})
+	actual, err := ioutil.ReadAll(tr)
+	require.NoError(t, err)
+	assert.Exactly(t, expected, string(actual))
+
+	_, err = r.Seek(0, io.SeekStart)
+	require.NoError(t, err)
+
+	// Partial parse
+	var partial [11]byte
+	tr = transform.NewReader(r, &expandTransformer{expand: expand_func})
+	n, err := tr.Read(partial[:])
+	require.NoError(t, err)
+	assert.Exactly(t, n, len(partial))
+	assert.Exactly(t, expected[:n], string(partial[:]))
+
+	// Empty string
+	r = bytes.NewReader([]byte{})
+	tr = transform.NewReader(r, &expandTransformer{expand: expand_func})
+	actual, err = ioutil.ReadAll(tr)
+	require.NoError(t, err)
+	assert.Exactly(t, "", string(actual))
+}
+
+func TestExpanderOneByteAtATime(t *testing.T) {
+	t.Parallel()
+
+	r := bytes.NewReader([]byte(orig))
+	rr := &oneByteReader{r: r}
+
+	expand_func := func(s string) (string, error) {
+		switch s {
+		case "t3sT":
+			return "test", nil
+		case "i5_":
+			return "is ", nil
+		}
+
+		return "NOMATCH", errors.New("No Match")
+	}
+
+	tr := transform.NewReader(rr, &expandTransformer{expand: expand_func})
+	actual, err := ioutil.ReadAll(tr)
+	require.NoError(t, err)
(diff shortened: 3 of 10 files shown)


alsamylkin:
Thank you for the PR Brandon! I really like your fix on expanding strings started with $.(for casting we recommend to use [spf13/cast library](https://github.com/uber-go/config/blob/master/example_cast_test.go#L31)) I'll do a more thorough review tomorrow morning, but there are couple small nits I see now:
1. Please update the changelog.
2. Could you add more tests for transform: strings that ends with `$`, `$$`, consist only of `$$$$$$$`, empty strings?

expand.go
```diff
@@ -0,0 +1,206 @@
...
+	"io"
+)
+
+// expandReadCloser implements io.ReadCloser
+type expandReadCloser struct {
+	tReader *transform.Reader
+	rc      io.ReadCloser
+}
+
+func newExpandReadCloser(rc io.ReadCloser, expand_func func(string) (string, error)) *expandReadCloser {
```
alsamylkin: please, use camelCase for arguments.

yaml_test.go
```diff
@@ -1745,6 +1748,31 @@ func TestNewYamlProviderWithExpand(t *testing.T) {
...
+	assert.IsType(t, int(12), p.Get("a-int").Value())
+	assert.Exactly(t, p.Get("a-int").Value(), pp.Get("a-int").Value())
+	assert.IsType(t, nil, p.Get("a-nil").Value())
+	assert.Exactly(t, p.Get("a-nil").Value(), pp.Get("a-nil").Value())
+	assert.IsType(t, nil, p.Get("a-null").Value())
+	assert.Exactly(t, p.Get("a-null").Value(), pp.Get("a-null").Value())
+	assert.IsType(t, "string", p.Get("a-string").Value())
+	assert.Exactly(t, p.Get("a-string").Value(), pp.Get("a-string").Value())
+}
+
```
alsamylkin: Could you add more tests for expand:
```golang
p, err := NewStaticProviderWithExpand(
	map[string]interface{}{
		"nil":   "a",
		"one":   "$a",
		"two":   "$$a",
		"three": "$$$a",
		"four":  "$$$$a",
		"five":  "$$$$$a",
	},
	func(key string) (string, bool) {
		return fmt.Sprint(len(key)), true
	})

require.NoError(t, err)
assert.Equal(t, "a", p.Get("nil").String())
assert.Equal(t, "1", p.Get("one").String())
assert.Equal(t, "$a", p.Get("two").String())
assert.Equal(t, "$1", p.Get("three").String())
assert.Equal(t, "$$a", p.Get("four").String())
assert.Equal(t, "$$1", p.Get("five").String())
```
 I am pretty sure, it will pass, but I want to keep them for future refactoring(if any) to be easy.

expand.go
drafnel: Will do.

yaml_test.go
drafnel: Thanks for these.

drafnel:
So how do I update my commits in github?  Can I just force push to the branch in my repo?

alsamylkin:
Yes, I think a regular push to your branch will be enough.

drafnel:
Cool, that worked.  Series updated.

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## v1.0.2 (unreleased)
```
breerly: Looks like we're introducing new functionality here? This should be `v1.1.0` in that case.

drafnel: Agreed.

I consider most of this to be either a tightening of the existing rules (e.g. rules for variable names) or a correction of less-than-desired behavior with respect to the underlying data types of Value objects that are produced when variable substitution is active.

The only part I'd really consider a new feature is the '$$' handling.

Any of that is enough to justify more than a patch-level version bump.

drafnel:
Noticed a bug after rereading the text.transform docs.  nil error should only be returned from the Transform function if all of the bytes in src were consumed.  Series updated to 300123e.  Here's what the interdiff looks like:

    $ git diff 32aa395 300123e
    diff --git a/expand.go b/expand.go
    index da068e0..4524827 100644
    --- a/expand.go
    +++ b/expand.go
    @@ -87,7 +87,12 @@ func (this *expandTransformer) Transform(dst, src []byte, atEOF bool) (int, int,
            var i int
            var j int

    -       for i < len(src) && j < len(dst) {
    +       for i < len(src) {
    +
    +               if j == len(dst) {
    +                       return j, i, transform.ErrShortDst
    +               }
    +
                    // Not a token-starting character, just copy it
                    if src[i] != '$' {
                            dst[j] = src[i]

drafnel:
Just a few last tweaks to prefer bytes.IndexByte() and copy() where possible, and simplify the way bare variable names are parsed.  I don't expect to make any more changes.  Here's the interdiff against the last update:

    diff --git a/expand.go b/expand.go
    index 4524827..2c03721 100644
    --- a/expand.go
    +++ b/expand.go
    @@ -70,10 +70,11 @@ func isShellNameChar(c byte) bool {
     		(c >= '0' && c <= '9')
     }
     
    -// bytesIndexFunc is what bytes.IndexFunc should be
    -func bytesIndexFunc(buf []byte, f func(b byte) bool) int {
    +// bytesIndexCFunc returns the index of the byte for which the
    +// complement of the supplied function is true
    +func bytesIndexCFunc(buf []byte, f func(b byte) bool) int {
     	for i := 0; i < len(buf); i++ {
    -		if f(buf[i]) {
    +		if !f(buf[i]) {
     			return i
     		}
     	}
    @@ -93,14 +94,29 @@ func (this *expandTransformer) Transform(dst, src []byte, atEOF bool) (int, int,
     			return j, i, transform.ErrShortDst
     		}
     
    -		// Not a token-starting character, just copy it
    -		if src[i] != '$' {
    -			dst[j] = src[i]
    -			i++
    -			j++
    +		end := bytes.IndexByte(src[i:], '$')
    +
    +		if end == -1 {
    +			// src does not contain '$', copy into dst
    +			cnt := copy(dst[j:], src[i:])
    +			i += cnt
    +			j += cnt
     			continue
    +		} else if end > 0 {
    +			// copy chars preceding '$' from src to dst
    +			cnt := copy(dst[j:], src[i:i+end])
    +			i += cnt
    +			j += cnt
    +
    +			if j == len(dst) {
    +				return j, i, transform.ErrShortDst
    +			}
    +
    +			// fallthrough
     		}
     
    +		// src[i] now points to '$', j < len(dst)
    +
     		// If we're at the end of src, but we found a starting
     		// token, return ErrShortSrc, unless we're also at EOF,
     		// in which case just copy it dst.
    @@ -126,7 +142,6 @@ func (this *expandTransformer) Transform(dst, src []byte, atEOF bool) (int, int,
     		}
     
     		var token []byte
    -		var end int
     
     		// Start of bracketed token ${foo}
     		if src[i+1] == '{' {
    @@ -168,9 +183,7 @@ func (this *expandTransformer) Transform(dst, src []byte, atEOF bool) (int, int,
     				continue
     			}
     
    -			end = bytesIndexFunc(src[i+2:], func(c byte) bool {
    -				return !isShellNameChar(c)
    -			})
    +			end = bytesIndexCFunc(src[i+2:], isShellNameChar)
     
     			if end == -1 {
     				// Reached the end of src without finding

drafnel:
Hey @alsamylkin is there anything holding this up?

expand_test.go
```diff
@@ -0,0 +1,244 @@
...
+	expand_func := func(s string) (string, error) {
+		switch s {
+		case "a":
+			return a, nil
+		}
+
+		return "NOMATCH", errors.New("No Match")
+	}
+
+	for _, tst := range tests {
```
alsamylkin: Could you use [subtests](https://blog.golang.org/subtests) instead of table driven tests?

expand.go
```diff
@@ -0,0 +1,224 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package config
+
+import (
+	"bytes"
+	"golang.org/x/text/transform"
```
alsamylkin: Looks like this import should be in a separate group.

drafnel: I couldn't care less.  So you're saying you want a blank line between these two imports?

glibsm:
```
import (
  // stdlib

 // current project

 // external
)
```

Do you have `go imports` configured for your editor?

expand.go
```diff
@@ -0,0 +1,224 @@
...
+func newExpandReadCloser(rc io.ReadCloser, expandFunc func(string) (string, error)) *expandReadCloser {
+	return &expandReadCloser{
+		tReader: transform.NewReader(
+			rc,
+			&expandTransformer{expand: expandFunc}),
+		closer: rc,
+	}
+}
+
+func (this *expandReadCloser) Read(p []byte) (int, error) {
```
drafnel: I consider that advice from the go project to be the worst kind of advice.  "Use single letter variable names".  Garbage.  When you see the letter 'b' in a function, should you immediately assume that it refers to the calling object?  What about the letter 'c'?  I often use the letter 'c' for a variable that holds a single character.  However, if you see a variable named "this", every developer will immediately understand that it refers to the calling object.

Now, I understand that "methods" in Go are much more similar to the way you do objects in C, by passing in the object as the first argument.  So while a Go method definition might look like this:

   func (*ObjectType) foo (...)

it is equivalent to a function defined like this:

   func foo (*ObjectType, ...) 

So what?  It does not improve readability or remove ambiguity by using a variable named 'o' to refer to the object which called this function.  Why Rob Pike thinks this is a good idea, and actively rails against using something that I consider both more readable and less ambiguous, is beyond me.

Python has the same calling convention for methods as Go does, i.e. the calling object is passed in as the first argument and can be given any name by the developer.  At least that project had the sense to understand that there is zero value in using a different name for the calling object in every different class, and there is a negative impact on readability, so they suggest calling it "self".  That's a convention worth documenting IMHO.

But, it's your project so I'll use whatever letter you want me to use.  I can't think of a single letter that would be immediately recognized by a random developer as referring to the calling object.  Please tell me which letter you'd like me to use.

expand_test.go
drafnel: Sigh...

expand.go
alsamylkin: I don't really care about the rule, some linters are using all the rules from the wiki to validate code(e.g. my VSCode) so I just want to keep them happy. Lets use `e` for the `expandReadCloser` struct.

expand.go
alsamylkin: @drafnel  yes just add a blank line. Somehow `expand_test.go` picked it up correctly, but `expand.go` didn't.

expand.go
```diff
@@ -0,0 +1,224 @@
...
+
+import (
+	"bytes"
+	"golang.org/x/text/transform"
+	"io"
+)
+
+// expandReadCloser implements io.ReadCloser
+type expandReadCloser struct {
+	tReader *transform.Reader
```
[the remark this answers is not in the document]
drafnel: This is exactly what I wanted to do in the first place, but I didn't know I could.  Thanks!

CHANGELOG.md
```diff
@@ -1,5 +1,24 @@
 # Changelog
 
+## v1.1.0 (unreleased)
+
+- Expand variables (e.g. $foo ${foo}) using a text.Transform instead
```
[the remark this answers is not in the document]
drafnel: Sure, I can add this detail to the NewYAMLProviderWithExpand docs.

expand_test.go
```diff
@@ -0,0 +1,244 @@
...
+	n := copy(buf, this.buf[this.offset:])
+	this.offset += n
+	if this.offset == len(this.buf) {
+		err = io.EOF
+	}
+	return n, err
+}
+
+func TestExpander(t *testing.T) {
+	var buf []byte
```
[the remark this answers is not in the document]
drafnel: Sure.

expand_test.go
drafnel: Sure.

expand.go
drafnel: @glibsm no I don't use `go imports`.  I prefer not to have tools rewriting my code behind my back.

akshayjshah: That's absolutely your prerogative, but you'll get less grief if you take a little time to internalize the formatting that it imposes. Many projects will require that your code conform to the same style.

alsamylkin:
@breerly, @akshayjshah are you ok with these changes?

drafnel:
> Most importantly, please sign the CLA!

Didn't I already sign this when I submitted to thriftrw?  Do I need to do this for every single repo?

> We can fix the unintended type changes by adding another case to the type switch in recursiveApply and only expanding strings. This has the added benefit of not making the final type of n: ${FOO} depend on the value of $FOO - it's always a string, which matches the current behavior (and is IMO more reasonable).

I don't find that behavior more reasonable.  Why should it be reasonable for a value of '13' to sometimes be a string and sometimes be an integer?

I would find it less surprising (and hence more reasonable) if '13' always produced an integer regardless of whether the field in the yaml file was specified like:

    foo: 13

or like:

    foo: ${FOO:13}

Once you have defined behavior that is also consistent, it allows the call-sites to be simpler.

> We can support the $-escaping, using exactly the rules you've proposed, by special-casing in == "$" in the anonymous function in replace. See https://play.golang.org/p/Rl2ywc3yfv for an example of this working.

Yup, that's true.  It's not necessarily the right place to put it depending on how you draw your boundaries of concern though.

What you don't get with either of your suggestions is an elegant solution in my opinion.  io.Reader is a natural interface for this and allows the code within yaml.go to be nicely simplified since the io.Reader provided by expand.go can be wrapped around the file readers and then passed to the lower level functions within yaml.go.  You have a net reduction in code paths since you don't have to have special cases for the "WithExpand" case.

Ideally, the yaml parser would accept an io.Reader too instead of requiring the caller to do an io.ReadAll() and then pass in the byte slice.  At some point I expect such a yaml parser will be created and yaml.go can be simplified further.

-Brandon

expand.go
```diff
@@ -0,0 +1,217 @@
...
+}
+
+// expandTransformer implements transform.Transformer
+type expandTransformer struct {
+	transform.NopResetter
+	expand func(string) (string, error)
+}
+
+// First char of shell variable may be [a-zA-Z_]
+func isShellNameFirstChar(c byte) bool {
```
abhinav: Looking at the implementation of os.Expand, it looks like it supports `$1` and `$1foo`. Should we replicate that behavior?

https://play.golang.org/p/3g_9wwuQU_

expand.go
```diff
@@ -0,0 +1,217 @@
...
+		(c >= 'A' && c <= 'Z') ||
+		(c >= '0' && c <= '9')
+}
+
+// bytesIndexCFunc returns the index of the byte for which the
+// complement of the supplied function is true
+func bytesIndexCFunc(buf []byte, f func(b byte) bool) int {
+	for i := 0; i < len(buf); i++ {
+		if !f(buf[i]) {
+			return i
```
abhinav: optional nit: why use this instead of `range`?

```
for i, b := range buf {
  if !f(b) {
    return i
  }
}
```

expand.go
```diff
@@ -0,0 +1,217 @@
...
+	}
+	return -1
+}
+
+// Transform expands shell-like sequences like $foo and ${foo} using
+// the configured expand function.  The sequence '$$' is replaced with
+// a literal '$'.
+func (e *expandTransformer) Transform(dst, src []byte, atEOF bool) (int, int, error) {
+	var i int
+	var j int
```
abhinav: nit: can we use the names used in the interface? https://godoc.org/golang.org/x/text/transform#Transformer

The names nDst and nSrc are more self-documenting than i and j.

expand.go
```diff
@@ -0,0 +1,217 @@
...
+			// copy chars preceding '$' from src to dst
+			cnt := copy(dst[j:], src[i:i+end])
+			i += cnt
+			j += cnt
+
+			if j == len(dst) {
+				return j, i, transform.ErrShortDst
+			}
+
+			// fallthrough
```
abhinav: comment not necessary. fallthrough is expected.

expand.go
```diff
@@ -0,0 +1,217 @@
...
+
+	for i < len(src) {
+
+		if j == len(dst) {
+			return j, i, transform.ErrShortDst
+		}
+
+		end := bytes.IndexByte(src[i:], '$')
+
+		if end == -1 {
```
abhinav: Nit: inline the declaration of `end` into the if-else block. It's not used
outside it.

```
if end := bytes.IndexByte(...); end == -1 {
  // ..
} else if end > 0 {
  // ..
}
```

The intention is to keep the scope of the variable as small as possible.

expand.go
```diff
@@ -0,0 +1,217 @@
...
+// the configured expand function.  The sequence '$$' is replaced with
+// a literal '$'.
+func (e *expandTransformer) Transform(dst, src []byte, atEOF bool) (int, int, error) {
+	var i int
+	var j int
+
+	for i < len(src) {
+
+		if j == len(dst) {
+			return j, i, transform.ErrShortDst
```
abhinav: This and the other "ErrShortDst/ErrShortSrc" cases don't have coverage. These look like important corner cases we should have tests for.

expand.go
```diff
@@ -0,0 +1,217 @@
...
+			j++
+			continue
+		}
+
+		var token []byte
+
+		// Start of bracketed token ${foo}
+		if src[i+1] == '{' {
+			end = bytes.IndexByte(src[i+2:], '}')
+			if end == -1 {
```
abhinav: Once the other `end` is out of scope, you should declare this one right above
this `if` block with a slightly more descriptive name.  Something like,
`varEnd` or `tokenEnd` maybe, or a comment to go with it.

```
var tokenEnd int  // end of the $foo or ${foo}
if ... {
  // ...
  tokenEnd = ...
} else {
  // ...
  tokenEnd = ...
}
```

It makes it much clearer that `tokenEnd` is the output of the following if-
else block.

expand.go
```diff
@@ -0,0 +1,217 @@
...
+			end = bytesIndexCFunc(src[i+2:], isShellNameChar)
+
+			if end == -1 {
+				// Reached the end of src without finding
+				// end of shell variable
+				if !atEOF {
+					// We need more bytes in src
+					return j, i, transform.ErrShortSrc
+				}
+				end = len(src)
```
abhinav: Can we add a test where `$foo` is at the end of the string (so at EOF). That'll cover this line.

expand.go
```diff
@@ -0,0 +1,217 @@
...
+
+			token = src[i+1 : end]
+		}
+
+		replacement, err := e.expand(string(token))
+		if err != nil {
+			return j, i, err
+		}
+
+		cnt := copy(dst[j:], replacement)
```
abhinav: Shouldn't you make sure there's enough room in `dst` before copying into it? If the replacement is too large, it'll be copied partially, then copied again when this is retried after `ErrShortDst` is returned.

expand.go
```diff
@@ -0,0 +1,217 @@
...
+		if cnt == len(replacement) {
+			i = end
+			j += cnt
+			continue
+		}
+
+		// If we couldn't copy the replacement string into dst,
+		// we need more room in dst.
+
+		return j, i, transform.ErrShortDst
```
abhinav: Can we test for this by providing a buffer that is too short for a very large replacement string?

expand.go
drafnel: Well, `$1foo` would be parsed as `$1` followed by the literal characters `foo`.  In shell context $1, $2, etc refer to parameter arguments,  $* refers to the concatenated parameter arguments, $@ refers to the quoted parameter arguments, $$ refers to the process id, etc.  They all have specific meaning in a shell context which don't really seem to make sense here.  I'd stick strictly to shell variable names which have the form "[a-zA-Z_][a-zA-Z0-9_]*".  But I don't feel too strongly about it.

Of course, the code I've uploaded doesn't strictly parse variables wrapped in curly braces, so it's possible that someone could use ${1} as a variable name.  Not that that's a good thing, just pointing it out.

abhinav: I don't feel strongly about whether we should support `$1` or not. It won't be a valid environment variable for sure. I'm happy to leave that as-is.

expand.go
drafnel: No reason, just didn't occur to me.  I don't feel strongly about either form, so I can use range if you prefer.

expand.go
drafnel: Sure.  I have a mild resistance to the 'n' prefix, which is generally reserved for holding a "count" or "number" of X, but I'll change it.

expand.go
drafnel: You may have been looking at an older version of this commit.  The current commit should have 100% test coverage of expand.go.  A couple of these were particularly difficult to trigger and require intimate knowledge of the text/transform package.

expand.go
drafnel: Already exists.

expand.go
abhinav: Yeah, please change it to range. Thanks.

expand.go
abhinav: I'm not committed to the naming as done by the Transformer interface. Similar variable names that are more descriptive would work too (srcPos, dstPos).

expand.go
drafnel: There should already be a test that triggers this code path.  Ultimately the text/transform package currently uses a buffer that is 4096 bytes, so a substitution may not be longer than that.

expand.go
drafnel: Yes, the comment is purely to communicate to the reader that the fallthrough is intentional in the 'else' block, unlike the 'if' portion of the block which calls continue.  Happy to delete it if it does not improve readability.

expand.go
drafnel: Sure.

expand.go
drafnel: Makes sense.

expand.go
drafnel: Hmm, I think I find

    if end := bytes.IndexByte(src[srcPos:], '$'); end == -1 {
        ...
    } else if end > 0 {

to be considerably less readable than

    end := bytes.IndexByte(src[srcPos:], '$')
    if end == -1 {
        ...
    } else if end > 0 {

Maybe the notation will grow on me.  For now, I'd prefer to leave it as is.

expand.go
drafnel: The copy itself was being used to check whether there was enough room.  I expect the normal case to be that there is enough room, but sure, we can the length and then avoid the unnecessary copy here and above.

drafnel:
Ok, I think I addressed all of the comments.  Let me know if I missed anything.  I'll hold off on resolving the merge until we decide how to handle pull request #64 which I'd like to slip in after the first commit in this series which would allow me to delete the expandReadCloser object from expand.go.

expand.go
abhinav: copy will copy at most `len(dst)` bytes, right? The issue I was pointing out was that in the case where `dst` is too short, copy will copy some number of bytes less than `len(src)`, we'll return `ErrDstShort`, in which case text/transform will retry after resizing dst, and we'll copy the part of `src` that was already copied again. (This all, assuming I understand the contract of text/transform correctly.)

drafnel: Yes that's correct.  I think I understood your original comment.  In my reply when I referred to "the unnecessary copy", I was talking about the case you described, the first copy that would happen when dst was too short, and which would then be performed again in full after dst was resized.

expand.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+	return -1
+}
+
+// Transform expands shell-like sequences like $foo and ${foo} using
+// the configured expand function.  The sequence '$$' is replaced with
+// a literal '$'.
+func (e *expandTransformer) Transform(dst, src []byte, atEOF bool) (int, int, error) {
+	var srcPos int
+	var dstPos int
```
abhinav: nit: you can use `var srcPos, dstPos int` here. The "type after names" syntax is there for exactly this reason.

expand.go
```diff
@@ -0,0 +1,213 @@
...
+		// sequence, emit a '$' into dst.
+		if src[srcPos+1] == '$' {
+			dst[dstPos] = src[srcPos]
+			srcPos += 2
+			dstPos++
+			continue
+		}
+
+		var token []byte
+		var tokenEnd int
```
abhinav: nit: var groups are cool

```go
var (
  token []byte
  tokenEnd int
)
```

expand.go
drafnel: I don't think that improves readability, so, perhaps with some exceptions, I tend to use one declaration per line.

expand.go
drafnel: Oh let me count the multitude of ways by which a variable can be defined.  I think Rob Pike forgot that your design isn't done when there's nothing left to add, it's done when there's nothing left to take away :-)

drafnel:
Ok, I inserted the commit from #64 into this series like I mentioned before.  If everyone's happy, I think this is ready to merge.

alsamylkin:
Please don't insert it, until the #64 is approved.
Both changes will be released together, no need to merge PR.

abhinav:
Regarding moving the changes into this PR:

What @drafnel intends for us to do here is to merge the PR without squashing.
If you look at the individual commits, they're all highly detailed commit
messages. If we merge the changes as-is without squashing, the PRs will be
closed as expected and the commits will tell a story similar to if @drafnel
created 5 different PRs.

I think the two PRs should be fine to merge together once #64 gets approved.

abhinav:
I'm merging these changes in. Writing tests for this specific case is coverage
for the sake of coverage and would require more refactoring than it's worth.

abhinav:
I'm merging these changes in. Writing tests for this specific case is coverage
for the sake of coverage and would require more refactoring than it's worth.

PR #60, opened 2017-08-14
Fixed populate panic for a nil pointer

alsamylkin:
We can't populate a value if there no value.
`reflect.Indirect` of a nil pointer returns an invalid `reflect.Value` and panics in the decoder.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,8 @@
 # Changelog
 
-## v1.1.0 (not released yet)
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,8 +1,8 @@
 # Changelog
 
-## v1.1.0 (not released yet)
+## v1.0.2 (2017-08-14)
```
amckinney: Update to 2017-08-16

Issue #61, opened 2017-08-25
android arm build overflow


ghost:
From https://github.com/uber-go/fx/issues/594
`GOOS=android GOARCH=arm CGO_ENABLED=0 go build` fails to compile:
```
./decoder.go:76: constant 9223372036854775807 overflows uintptr
./decoder.go:122: constant 18446744073709551615 overflows uintptr
```

akshayjshah:
We're not planning to fix this; if we ever need to deploy on ARM and/or Android, we can look into it.

PR #62, opened 2017-08-25
Convert uintptr to int64/uint64 to compare to max constants

alsamylkin:
Golang has a different pointer size for arm/x86(4) and x64(8): https://github.com/golang/debug/blob/master/arch/arch.go#L160.
Which leads to a compile error when comparing these types with constant expressions `math.MaxInt64`/ `math.MaxUInt64`.

Fixes https://github.com/uber-go/config/issues/61.

### DIFF
--- a/decoder.go
+++ b/decoder.go
@@ -73,7 +73,7 @@ func convertSignedInts(src interface{}, dst *reflect.Value) error {
 			return nil
 		}
 	case uintptr:
-		if t <= math.MaxInt64 && !dst.OverflowInt(int64(t)) {
+		if int64(t) <= math.MaxInt64 && !dst.OverflowInt(int64(t)) {
 			dst.SetInt(int64(t))
 			return nil
 		}
@@ -119,7 +119,7 @@ func convertUnsignedInts(src interface{}, dst *reflect.Value) error {
 			return nil
 		}
 	case uintptr:
-		if t <= math.MaxUint64 && !dst.OverflowUint(uint64(t)) {
+		if uint64(t) <= math.MaxUint64 && !dst.OverflowUint(uint64(t)) {
 			dst.SetUint(uint64(t))
 			return nil
 		}
--- a/decoder.go
+++ b/decoder.go
@@ -57,7 +57,7 @@ func derefType(t reflect.Type) reflect.Type {
 
 func convertSignedInts(src interface{}, dst *reflect.Value) error {
 	switch t := src.(type) {
-	case int, uint, int8, uint8, int16, uint16, int32, uint32, int64:
+	case int, uint, int8, uint8, int16, uint16, int32, uint32, int64, uint64, uintptr:
 		i, err := strconv.ParseInt(fmt.Sprint(t), 10, 64)
 		if err != nil {
 			return err
@@ -65,16 +65,6 @@ func convertSignedInts(src interface{}, dst *reflect.Value) error {
 
 		if !dst.OverflowInt(i) {
 			dst.SetInt(i)
-			return nil
-		}
-	case uint64:
-		if t <= math.MaxInt64 {
-			dst.SetInt(int64(t))
-			return nil
-		}
-	case uintptr:
-		if int64(t) <= math.MaxInt64 && !dst.OverflowInt(int64(t)) {
-			dst.SetInt(int64(t))
 			return nil
 		}
 	case float32:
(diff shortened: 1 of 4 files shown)


alsamylkin:
Not sure, if we should add an extra target for CI, but this one works locally.
image

decoder.go
```diff
@@ -73,7 +73,7 @@ func convertSignedInts(src interface{}, dst *reflect.Value) error {
 			return nil
 		}
 	case uintptr:
-		if t <= math.MaxInt64 && !dst.OverflowInt(int64(t)) {
+		if int64(t) <= math.MaxInt64 && !dst.OverflowInt(int64(t)) {
```
prashantv: Why do you need both `<= math.MaxInt64` and `dst.OverflowInt` check?

prashantv: should we use `OverflowUint` for `uintptr`?

alsamylkin: The `dst` could represent any bit sized integer, and `Value.SetInt` will use [conversion](https://golang.org/src/reflect/value.go?s=42937:42967#L1426) instead of reporting overflows.

alsamylkin: This method is for setting `dst`, which is an integer in this method. `OverflowUint` will panic in this case.

alsamylkin: There is another bug in the code though,  int64(math.MaxUint64) will be -1 :)

prashantv: `int64(t)` can only ever be a value that is `math.MinInt64 <= t <= math.MaxInt64` (by definition, since you're casting it to an `int64`), so the check `int64(t) <= math.MaxInt64` will *always* be true.

drafnel: If you cast t to int64 it will necessarily be <= math.MaxInt64.  So the comparison `int64(t) <= math.MaxInt64` will always be true right?  I think what you need to do here is cast to an unsigned type and then do your comparison.  Maybe something like:

    if uint64(t) <= math.MaxInt64 ...

Also, just a question.  Why, in the first case block in this switch, do you convert the integer value 't' to a string and then parse it again?  Can't you just cast 't' to int64, check for overflow and then call SetInt()?

And why aren't you checking for overflow in the uint64 case?

decoder.go
```diff
@@ -119,7 +119,7 @@ func convertUnsignedInts(src interface{}, dst *reflect.Value) error {
 			return nil
 		}
 	case uintptr:
-		if t <= math.MaxUint64 && !dst.OverflowUint(uint64(t)) {
+		if uint64(t) <= math.MaxUint64 && !dst.OverflowUint(uint64(t)) {
```
drafnel: Ditto here.  Casting to uint64 makes the comparison moot.  Until there is a datatype larger than uint64 that uintptr could have and that you could cast both sides to, I don't think this comparison provides any value.

decoder.go
alsamylkin: Yeah :) I am trying to find a max of uintptr, but can't think of any reasonable approach. Any suggestions?

prashantv: You have a `uintptr`, and you want to verify it's safe to write in to a signed integer right?

If you convert to a `int64`, and it's negative, that means the value is too large for any signed integer. If it's positive, but too large for the specific signed integer type, `OverflowInt` will catch it.

So I think what you want is:
```
if int64(t) >= 0 && !dst.OverflowInt(int64(t)) {
```

alsamylkin: Well, it is a signed integer here, but an unsigned integer below. Probably, we can assume, that uintptr is always smaller than math.MaxUInt64 and report an error to the user otherwise.

alsamylkin: I am using Print-Parse pair, because in case of multiple types in the type switch, `t` is going to be an `interface{}` and I can't use it to call `Value.SetInt`.
So either this or have multiple small `case`s for each int8, 16, ...

drafnel: Ah.  I'm unfamiliar with the nuances of type assertion switches.  I think you should think about  going with the small case statements and avoid running the src value through print/parse.

alsamylkin: I decided to keep it simple and use keep uintptr in the same bay as other ints.

decoder.go
alsamylkin: Yeah, late night PR's should be prohibited :)

decoder.go
alsamylkin: Well, that was my initial implementation :) I've received a comment about bad readability and put all integer numbers together. We already doing lots of reflection, which always allocates on heap, so there is not much perf gain we can receive.

drafnel: Never be afraid to reject bad advice.  I think it's good to always try to think of your code and write your code as if it was an example of what to do for other engineers and assume they will copy what you do.  In other words, everything you write is a teaching opportunity.  I don't think you can say that about this print/parse integer code.  Some engineer will come along and see this and say "Hey, that's a neat trick, I should use that", without realizing the implications.

alsamylkin: I am happy to chat about these improvements offline since it feels like an off topic :)

PR #64, opened 2017-08-29
Bc/no leak file handles

drafnel:
I noticed the other day that file handles are not closed on error, and so file descriptors may be leaked.

May want to slip this patch into the "Perform variable expansion" series (#58) in between commit 78fd180 "yaml.go use newYAMLProviderFromReader() in a couple more places" and commit 0e221a0 "Demonstrate that YAML providers produce different data types", regardless of what is decided about the last commit in that series.

-Brandon

### DIFF
--- a/yaml.go
+++ b/yaml.go
@@ -44,7 +44,7 @@ var (
 	_emptyDefault = `""`
 )
 
-func newYAMLProviderCore(files ...io.ReadCloser) (*yamlConfigProvider, error) {
+func newYAMLProviderCore(files ...io.Reader) (*yamlConfigProvider, error) {
 	var root interface{}
 	for _, v := range files {
 		var curr interface{}
@@ -134,28 +134,57 @@ func mergeMaps(dst interface{}, src interface{}) (interface{}, error) {
 // file names. All the objects are going to be merged and arrays/values
 // overridden in the order of the files.
 func NewYAMLProviderFromFiles(files ...string) (Provider, error) {
-	readers, err := filesToReaders(files...)
+	readClosers, err := filesToReaders(files...)
 	if err != nil {
 		return nil, err
 	}
 
-	return newYAMLProviderFromReader(readers...)
+	readers := make([]io.Reader, len(readClosers))
+	for i, r := range readClosers {
+		readers[i] = r
+	}
+
+	provider, err := newYAMLProviderFromReader(readers...)
+
+	for _, r := range readClosers {
+		nerr := r.Close()
+		if err == nil {
+			err = nerr
+		}
+	}
+
+	return provider, err
 }
 
 // NewYAMLProviderWithExpand creates a configuration provider from a set of YAML
 // file names with ${var} or $var values replaced based on the mapping function.
 func NewYAMLProviderWithExpand(mapping func(string) (string, bool), files ...string) (Provider, error) {
-	readers, err := filesToReaders(files...)
+	readClosers, err := filesToReaders(files...)
 	if err != nil {
 		return nil, err
 	}
 
-	return newYAMLProviderFromReaderWithExpand(mapping, readers...)
+	readers := make([]io.Reader, len(readClosers))
+	for i, r := range readClosers {
+		readers[i] = r
+	}
+
+	provider, err := newYAMLProviderFromReaderWithExpand(mapping,
+		readers...)
+
+	for _, r := range readClosers {
+		nerr := r.Close()
+		if err == nil {
+			err = nerr
+		}
+	}
+
+	return provider, err
 }
 
-// NewYAMLProviderFromReader creates a configuration provider from a list of io.ReadClosers.
+// NewYAMLProviderFromReader creates a configuration provider from a list of io.Readers.
 // As above, all the objects are going to be merged and arrays/values overridden in the order of the files.
-func newYAMLProviderFromReader(readers ...io.ReadCloser) (Provider, error) {
+func newYAMLProviderFromReader(readers ...io.Reader) (Provider, error) {
 	p, err := newYAMLProviderCore(readers...)
 	if err != nil {
 		return nil, err
@@ -165,11 +194,11 @@ func newYAMLProviderFromReader(readers ...io.ReadCloser) (Provider, error) {
 }
 
 // NewYAMLProviderFromReaderWithExpand creates a configuration provider from
-// a list of `io.ReadClosers and uses the mapping function to expand values
+// a list of `io.Readers and uses the mapping function to expand values
 // in the underlying provider.
 func newYAMLProviderFromReaderWithExpand(
 	mapping func(string) (string, bool),
-	readers ...io.ReadCloser) (Provider, error) {
+	readers ...io.Reader) (Provider, error) {
 	p, err := newYAMLProviderCore(readers...)
 	if err != nil {
 		return nil, err
@@ -186,12 +215,12 @@ func newYAMLProviderFromReaderWithExpand(
 // blobs. As above, all the objects are going to be merged and arrays/values
 // overridden in the order of the yamls.
 func NewYAMLProviderFromBytes(yamls ...[]byte) (Provider, error) {
-	closers := make([]io.ReadCloser, len(yamls))
+	readers := make([]io.Reader, len(yamls))
 	for i, yml := range yamls {
-		closers[i] = ioutil.NopCloser(bytes.NewReader(yml))
+		readers[i] = bytes.NewReader(yml)
 	}
 
-	return newYAMLProviderFromReader(closers...)
+	return newYAMLProviderFromReader(readers...)
 }
 
 func filesToReaders(files ...string) ([]io.ReadCloser, error) {
@@ -200,6 +229,9 @@ func filesToReaders(files ...string) ([]io.ReadCloser, error) {
 
 	for _, v := range files {
 		if reader, err := os.Open(v); err != nil {
+			for _, r := range readers {
+				r.Close()
+			}
 			return nil, err
 		} else if reader != nil {
 			readers = append(readers, reader)
@@ -363,17 +395,13 @@ func (n *yamlNode) applyOnAllNodes(expand func(string) string) (err error) {
 	return
 }
 
-func unmarshalYAMLValue(reader io.ReadCloser, value interface{}) error {
+func unmarshalYAMLValue(reader io.Reader, value interface{}) error {
 	raw, err := ioutil.ReadAll(reader)
 	if err != nil {
 		return errors.Wrap(err, "failed to read the yaml config")
 	}
 
-	if err = yaml.Unmarshal(raw, value); err != nil {
-		return err
-	}
-
-	return reader.Close()
+	return yaml.Unmarshal(raw, value)
 }
 
 // Function to expand environment variables in returned values that have form: ${ENV_VAR:DEFAULT_VALUE}.
--- a/yaml.go
+++ b/yaml.go
@@ -139,12 +139,7 @@ func NewYAMLProviderFromFiles(files ...string) (Provider, error) {
 		return nil, err
 	}
 
-	p, err := newYAMLProviderCore(readers...)
-	if err != nil {
-		return nil, err
-	}
-
-	return newCachedProvider(p)
+	return newYAMLProviderFromReader(readers...)
 }
 
 // NewYAMLProviderWithExpand creates a configuration provider from a set of YAML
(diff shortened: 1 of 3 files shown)


drafnel:
Forgot to close the already-opened file handles in filesToReaders on failure.  Tweaked TestYAMLProviderWithGarbledPath to exercise this code path in the tests.

Here's the interdiff:

    diff --git a/yaml.go b/yaml.go
    index 42d3496..bb1eae4 100644
    --- a/yaml.go
    +++ b/yaml.go
    @@ -223,6 +223,9 @@ func filesToReaders(files ...string) ([]io.ReadCloser, error) {
     
     	for _, v := range files {
     		if reader, err := os.Open(v); err != nil {
    +			for j := 0; j < len(readers); j++ {
    +				readers[j].Close()
    +			}
     			return nil, err
     		} else if reader != nil {
     			readers = append(readers, reader)
    diff --git a/yaml_test.go b/yaml_test.go
    index dcda3c0..9cc5bcb 100644
    --- a/yaml_test.go
    +++ b/yaml_test.go
    @@ -1843,13 +1843,15 @@ func TestYAMLProviderWithGarbledPath(t *testing.T) {
     	t.Parallel()
     
     	t.Run("regular", func(t *testing.T) {
    -		_, err := NewYAMLProviderFromFiles("/some/nonexisting/config")
    +		_, err := NewYAMLProviderFromFiles("./testdata/base.yaml",
    +			"/some/nonexisting/config")
     		require.Error(t, err)
     		assert.Contains(t, err.Error(), "no such file or directory")
     	})
     
     	t.Run("expand", func(t *testing.T) {
    -		_, err := NewYAMLProviderWithExpand(nil, "/some/nonexisting/config")
    +		_, err := NewYAMLProviderWithExpand(nil, "./testdata/base.yaml",
    +			"/some/nonexisting/config")
     		require.Error(t, err)
     		assert.Contains(t, err.Error(), "no such file or directory")
     	})

yaml.go
```diff
@@ -134,33 +134,51 @@ func mergeMaps(dst interface{}, src interface{}) (interface{}, error) {
...
+	readers := make([]io.Reader, len(readclosers))
+	for i, r := range readclosers {
+		readers[i] = r
 	}
 
-	return newCachedProvider(p)
+	provider, err := newYAMLProviderFromReader(readers...)
+
+	for _, r := range readclosers {
+		r.Close()
```
alsamylkin: this [Close()](https://golang.org/pkg/io/#Closer) produces an error, which you skipped.

yaml.go
```diff
@@ -134,33 +134,51 @@ func mergeMaps(dst interface{}, src interface{}) (interface{}, error) {
...
+	readers := make([]io.Reader, len(readclosers))
+	for i, r := range readclosers {
+		readers[i] = r
+	}
+
+	provider, err := newYAMLProviderFromReaderWithExpand(mapping,
+		readers...)
+
+	for _, r := range readclosers {
+		r.Close()
```
alsamylkin: Same as above, we need to report errors on closers as well.

alsamylkin:
The tests in the patch above don't actually check that Close wasn't called..

yaml.go
drafnel: I'm not sure what error you expect to receive here.  If the files were opened for writing, then sure, the error is important since it could indicate that the write failed, but for reading?  The important error is the one returned by newYAMLProviderFromReader() above.  If that function returns an error, then I would think that would be the most important error to return, and the rest can be ignored.  If it does not return an error, meaning that all files were completely read and parsed successfully, then what value is there in returning any error from these calls to Close(), which I still can't think of how that would happen.

Even so, I'd be inclined to return the first error, and not try to concatenate all of the errors that occurred.  So something like:

    for _, r := range readclosers {
        e := r.Close()
        if err == nil {
            err = e
        }

drafnel:
I don't actually find that simpler, sorry :-/  The majority of these functions do not require a ReadCloser, they only really care about reading.  By passing them a ReadCloser, the responsibility for calling Close is being pushed deep down into functions that should not naturally have to worry about it.  And that's why there are so many places where a Reader is wrapped with io.NopCloser(), and it's why I had to make an expandReadCloser in expand.go.  The functions that do naturally have to worry about calling Close() and which should be responsible for calling Close() are the ones that are passed a filename and actually open the files.  Those functions open the files, and they should be responsible for closing them.

I know it looks like there are a lot of changes, but I think changing the interfaces to use a Reader instead of a ReadCloser everywhere produces a result that is a bit simpler than before and places the responsibility for calling Close() strictly in filesToReaders, NewYAMLProviderFromFiles, and NewYAMLProviderWithExpand where it belongs.  It will also allow us to delete the expandReadCloser in expand.go.

re: the tests.  The only real change in the tests is the change to TestYAMLProviderWithGarbledPath which merely ensures that the Close() code path is exercised in filesToReaders().

yaml.go
alsamylkin: Well, the most common case for errors from Close is that the file is already closed, which can't happen here, but there is also a case when the system call returned an error(EINTR). Which probably never happens in practice, but I still prefer to just to handle it and forget(how - doesn't really matter, aggregation is the simplest thing that came to my mind)

alsamylkin:
Removing `ioutil.NopCloser` in tests is ok. The reasons why I suggested a different approach are:
1. `Close` happens in one place, no need to duplicate the code.
2. Extra allocation: you create a copy of `io.Reader` slice, just to pass it around in varargs. It doesn't look very elegant to me.
3. Easier to test: right now tests check only for one error, that happens if e.g. a file is not found. The PR addresses an issue about closing all the files even if there was an error processing one of them and we don't actually verify that `Close` was ever called.

drafnel:
I think the important question to think about here is what is the right interface for these functions like newYAMLProviderCore, newYAMLProviderFromReader, unmarshalYAMLValue, etc.  My opinion is that these functions do not strictly _need_ to have an io.ReadCloser passed to them.  They are primarily interested in reading.  If we pass them an io.ReadCloser, then that implies that we are also transferring the responsibility for closing the object that we pass to them, but they don't inherently have a need to call Close().  So we can reduce the scope of responsibilities for these functions, which, as a side-effect, can make them easier to use, as evidenced by the numerous places where io.NopCloser no longer needs to be used.

> Close happens in one place, no need to duplicate the code.

That's a fine goal that I consider to be secondary to intuitive interfaces and readable code.  I just mean that I'm going to try to choose the right interfaces and try to choose the most readable solution even if that means that the same operations may exist in more than one place in the code.

It's always possible to avoid repetition if you try hard enough, and it would be possible to do so in the solution I proposed too, but I'm not sure it would result in code that is more readable.

> Extra allocation: you create a copy of io.Reader slice, just to pass it around in varargs. It doesn't look very elegant to me.

Well, we are at the mercy of the Go language.  It would be nice if it was possible to pass an []io.ReadCloser to a function that expects an []io.Reader, but it's not.  That being said, I don't think it's _that_ ugly.  It's just what you have to do in order to pass a slice of one interface type to a function that expects a slice of a different, but even though compatible, interface type.

The result here is that the interface and responsibilities of 8 functions has been simplified while slightly complicating the body of 2 functions.  It's a tradeoff.  But, in my opinion, those 2 functions should rightly have the responsibility for closing the files that they've opened, so I think it makes sense.

> Easier to test: right now tests check only for one error, that happens if e.g. a file is not found. The PR addresses an issue about closing all the files even if there was an error processing one of them and we don't actually verify that Close was ever called.

My feelings towards tests are similar to my feelings towards DRY.  It's a great principle to work towards, but I'm not going to bend over backwards and build multiple complicated levels of abstraction to comply with it.  So we /could/ create an additional function that NewYAMLProviderWithExpand and NewYAMLProviderFromFiles could call, that would accept an []io.ReadCloser, and then use that function in tests so that we could pass in our own mocked ReadCloser that we could use to verify that Close() was actually called, but I'm not sure it's worth the effort and will result in additional unnecessary complexity.  And then you'd still be left unable to test whether the functions that accept a filename _actually_ call Close() when they're supposed to.  Sometimes it's enough to just read the code and tell that yes, x-y-z will happen when it should.

So, I'd say give it some thought and think about what the right interfaces should be for these functions, maybe bounce it off some other people, and let me know what you think.

yaml.go
abhinav: It's good practice to handle the error unless you have a very specific reason for ignoring the error. Plus the contract for `io.Closer` doesn't let you make any assumptions about the error.

You can use [go.uber.org/multierr](https://godoc.org/go.uber.org/multierr) to do the concatenation for you without loss of information.

```go
var err error
for _, r := range readclosers {
  err = multierr.Append(err, r.Close())
}
```

yaml.go
```diff
@@ -134,33 +134,51 @@ func mergeMaps(dst interface{}, src interface{}) (interface{}, error) {
 // file names. All the objects are going to be merged and arrays/values
 // overridden in the order of the files.
 func NewYAMLProviderFromFiles(files ...string) (Provider, error) {
-	readers, err := filesToReaders(files...)
+	readclosers, err := filesToReaders(files...)
```
abhinav: nitty nit: readClosers

yaml.go
```diff
@@ -210,6 +223,9 @@ func filesToReaders(files ...string) ([]io.ReadCloser, error) {
 
 	for _, v := range files {
 		if reader, err := os.Open(v); err != nil {
+			for j := 0; j < len(readers); j++ {
+				readers[j].Close()
```
abhinav: You should handle the errors here too

yaml.go
```diff
@@ -383,7 +399,7 @@ func unmarshalYAMLValue(reader io.ReadCloser, value interface{}) error {
 		return err
 	}
 
-	return reader.Close()
+	return nil
```
abhinav: Can reduce this and the previous block to, `yaml.Unmarshal(raw, value)`.

drafnel:
Is it really impossible to _reply_ to github comments?  Or is there something I am missing?  I've been copy/pasting text every time to reply.

> It's good practice to handle the error unless you have a very specific reason for ignoring the error. Plus the contract for io.Closer doesn't let you make any assumptions about the error.

In general, I agree 100%.  Check errors.

I think the point here is that these errors are don't cares.  We've either already encountered an error while parsing the files, in which case we'd return that most-important first error, or we have successfully read and parsed all files, in which case the primary function of these calls has been successful and it is _possible_ to return the information that the caller requested, and it is _possible_ for the caller to continue.  If we return an error, then that indicates to the caller that the call has failed and that they should _not_ continue.  In general, I'd probably emit a warning here at most, or try to handle the error cases and retry the close if possible.

I don't think it's important enough to take a stand on, so I can collect the errors in a multierr and return it, regardless of how valueless I may think it is in this case :-)

drafnel:
Grr, multierr isn't currently used by this package.

abhinav:
Ah, feel free to add the dependency or just skip the error concatenation altogether. It's not a blocker either way.

drafnel:
Ok, I modified it so that the first encountered error will be returned.

alsamylkin:
> I think the important question to think about here is what is the right interface for these functions like newYAMLProviderCore, newYAMLProviderFromReader, unmarshalYAMLValue, etc. My opinion is that these functions do not strictly need to have an io.ReadCloser passed to them.

These are internal helper functions, we don't really care if they use `Reader` or `ReadCloser`, but if the latter allow us to remove duplication and improve testability, then why not to use it?

I will not approve the PR until there are tests covering the bug this PR addresses.

abhinav:
> These are internal helper functions, we don't really care if they use Reader or ReadCloser, but if the latter allow us to remove duplication and improve testability, then why not to use it?

@alsamylkin, it appears that there's a disagreement here over two approaches
to solving the same problem.

Your approach will leave most of the code unchanged, so it's simple in that it
doesn't touch a lot of code. However, from a design standpoint, @drafnel's
change is slightly safer and more future-proof because it exposes as little
power to the underlying layers as possible (see [Principle of least privilege][1])
at the cost of extra work for him. Since he has already done this work, I see
no reason to have him undo it for an easier approach.

[1]: https://en.wikipedia.org/wiki/Principle_of_least_privilege

> I will not approve the PR until there are tests covering the bug this PR addresses.

Looking at the code, there's no way to trigger `Close()` errors, or to assert
that `Close()` was even called without refactoring how files are opened. I
don't think that that level of refactoring should be part of this PR.

I would argue that this PR is good as-is.

alsamylkin:
@abhinav I don't care about using my approach, only about the bug fix to have a regression test. I am sorry, that my comments were understood that way: I showed one of the ways to deal with it and happy to accept any other code **with tests**.

Principle of the least privilege doesn't apply here: this is our internal file and there are no layers exposed, just our internal functions. Duplicating code actually increases probability of an error, especially when tests are harder to write.

abhinav:
I'm merging these changes in. Writing tests for this specific case is coverage
for the sake of coverage and would require more refactoring than it's worth.

PR #68, opened 2017-09-13
add support for loading config from bytes with expand

meiliang86:


akshayjshah:
Unfortunately, this is going to conflict with #58, which fixes a number of bugs in environment variable expansion.

I agree that this change is worthwhile, though - can you rebase and reimplement?

PR #77, opened 2018-05-23
Update dependencies

akshayjshah:
This is the first step in a series of config refactorings.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,3 @@ ci: SHELL := /bin/bash
 ci:
 	go test -race $(PACKAGES) -coverprofile=coverage.txt -covermode=atomic -count 2
 	bash <(curl -s https://codecov.io/bash)
-
-.PHONY: license
-license:
-	$(ECHO_V)./.build/license.sh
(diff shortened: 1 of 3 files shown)


glide.yaml
```diff
@@ -1,20 +1,20 @@
 package: go.uber.org/config
 import:
-- package: gopkg.in/validator.v2
-- package: gopkg.in/yaml.v2
 - package: github.com/pkg/errors
   version: ~0.8.0
 - package: golang.org/x/text
+  version: ~0.3.0
```
amckinney: Seeing as this is pre-1.0, this could potentially cause version mismatch errors from other libraries. We should take note of this in case it becomes an issue later.

Makefile
```diff
@@ -14,7 +14,3 @@ ci: SHELL := /bin/bash
 ci:
 	go test -race $(PACKAGES) -coverprofile=coverage.txt -covermode=atomic -count 2
 	bash <(curl -s https://codecov.io/bash)
-
-.PHONY: license
-license:
-	$(ECHO_V)./.build/license.sh
```
amckinney: No need for this anymore?

akshayjshah: Nope. It's wrong anyways.

Issue #76, opened 2018-05-15
Populate does not return a copy of the value, allowing callers to modify the config


prashantv:
See example test:
https://play.golang.org/p/c1LG6FyTBp2

Caller 1: Create and populate a map. Modify the map.
Caller 2: Calls Populate, sees values as modified by caller 1

```
=== RUN   TestProviderMutability
--- FAIL: TestProviderMutability (0.00s)
        Error Trace:    mutable_test.go:36
	Error:		Not equal: map[string]interface {}{"app":map[interface {}]interface {}{"foo":"bar"}} (expected)
			        != map[string]interface {}{"app":map[interface {}]interface {}{"foo":"baz"}} (actual)

			Diff:
			--- Expected
			+++ Actual
			@@ -2,3 +2,3 @@
			  (string) (len=3) "app": (map[interface {}]interface {}) (len=1) {
			-  (string) (len=3) "foo": (string) (len=3) "bar"
			+  (string) (len=3) "foo": (string) (len=3) "baz"
			  }

FAIL
exit status 1
```

akshayjshah:
Fixed in v1.2.0.

Issue #75, opened 2018-05-09
Cannot populate fields with omitempty tag


akshayjshah:
Fields tagged with both a name and the `omitempty` tag, e.g.

```
type Foo struct {
  Name string `yaml:"name,omitempty"`
}
```

aren't populated.

akshayjshah:
Fixed in v1.2.0.

Issue #72, opened 2017-12-15
YAML unmarshal seems to ignore null


prashantv:
Default values set using `WithDefault` are used even if a YAML provider tries to null out the value using yaml `~` syntax.  

Example code:
```go
        cfgP, err := config.NewYAMLProviderFromBytes([]byte(`
subkey: ~
`))
        if err != nil {
                panic(err)
        }

        type config struct {
                Initial   int `yaml:"initial"`
                Frequency int `yaml:"frequency"`
        }

        v, err := cfgP.Get("subkey").WithDefault(config{1000, 2000})
        if err != nil {
                panic(err)
        }

        var c config
        if err := v.Populate(&c); err != nil {
                panic(err)
        }

        fmt.Printf("%+v\n", c)
}
```

Expected output was to have empty config, but we get:
```
{Initial:1000 Frequency:2000}
```

In comparison, a YAML snippet which does something similar:
```go
package main

import (
        "fmt"

        yaml "gopkg.in/yaml.v2"
)

func main() {
        type subkey struct {
                Initial   int `yaml:"initial"`
                Frequency int `yaml:"frequency"`
        }
        type config struct {
                Subkey *subkey `yaml:"subkey"`
        }

        bs := []byte(`
subkey: ~
`)

        c := config{&subkey{1000, 2000}}
        if err := yaml.Unmarshal(bs, &c); err != nil {
                panic(err)
        }

        fmt.Println(c.Subkey)
}
```

Prints out `&{0 0}`.

akshayjshah:
Fixed in v1.2.0.

Issue #59, opened 2017-08-14
Add support for handling config structs' dependency (tag "yaml inline")


luozhaoyu:
It would duplicate a lot if several config structs are sharing the same set of fields, so that support for struct embedding would be very useful.

One solution is to add support for yaml tag "inline", which will make these codes works:

```go
package main

import (
	"fmt"
	"log"

	"go.uber.org/config"
)

type Nested struct {
	N string
}

func main() {
	p, _ := config.NewStaticProvider(map[string]interface{}{
		"example": map[string]interface{}{
			"name":    "uber",
			"founded": 2009,
			"n":       "nnnn",
		},
	})

	var c struct {
		Nested  `yaml:",inline"`
		Name    string
		Founded int
	}

	if err := p.Get("example").Populate(&c); err != nil {
		log.Fatal(err)
	}

	fmt.Println(c)
}
```

glibsm:
Thanks for the suggestion @luozhaoyu, we're going to look into it.

I took the liberty to edit the code in issue to make sure it runs and compiles with the latest version of the config library (import path `go.uber.org/config`).

hsuber:
In the mean time, can we please make the config loader fail fast if it encounters an 'inline' flag?  At least it wasn't clear to me and I spent a chunk of time debugging.

akshayjshah:
Fixed in v1.2.0.

PR #78, opened 2018-05-24
Add a suite of external integration tests

akshayjshah:
Before making any significant changes to the config logic, add a suite
of unit tests that exercise the externally-visible behavior of the
current implementation. A number of these tests currently fail, so
they're skipped and marked with todos.

This includes tests that replicate the following issues:

- https://github.com/uber-go/config/issues/72
- https://github.com/uber-go/config/issues/59
- https://github.com/uber-go/config/issues/76
- https://github.com/uber-go/config/issues/75

### DIFF
--- a/config_ext_test.go
+++ b/config_ext_test.go
@@ -39,7 +39,11 @@ fun:
 practical: &ptr
   toyota: camry
   honda: accord
-antique: model_t
+antique_scalar: model_t
+antique_sequence:
+  - model_t
+antique_mapping:
+  ford: model_t
 occupants:
   honda:
     driver: jane
@@ -56,7 +60,9 @@ fun:
 practical:
   honda: civic
   nissan: altima
-antique: ~
+antique_scalar: ~
+antique_sequence: ~
+antique_mapping: ~
 occupants:
   honda:
     passenger: arthur
@@ -76,7 +82,7 @@ mismatch: [scalar]
 		Populate  interface{}
 	}
 
-	run := func(tt TestCase) {
+	run := func(t testing.TB, tt TestCase) {
 		assert.Equal(t, tt.HasValue, tt.Value.HasValue(), "unexpected result from HasValue")
 		assert.Equal(t, tt.Interface, tt.Value.Value(), "unexpected result from Value")
 		require.NoError(t, tt.Value.Populate(tt.Populate), "error populating")
@@ -86,7 +92,7 @@ mismatch: [scalar]
 		var s string
 
 		t.Run("top level", func(t *testing.T) {
-			run(TestCase{
+			run(t, TestCase{
 				Value:     provider.Get("not_a_key"),
 				HasValue:  false,
 				Interface: nil,
@@ -96,7 +102,7 @@ mismatch: [scalar]
 		})
 
 		t.Run("subkey", func(t *testing.T) {
-			run(TestCase{
+			run(t, TestCase{
 				Value:     provider.Get("practical.cadillac"),
 				HasValue:  false,
 				Interface: nil,
@@ -106,7 +112,7 @@ mismatch: [scalar]
 		})
 
 		t.Run("descend into slice", func(t *testing.T) {
-			run(TestCase{
+			run(t, TestCase{
 				Value:     provider.Get("fun.extra"),
 				HasValue:  false,
 				Interface: nil,
@@ -118,7 +124,7 @@ mismatch: [scalar]
 
 	t.Run("nil", func(t *testing.T) {
 		var s string
-		run(TestCase{
+		run(t, TestCase{
 			Value:     provider.Get("nothing"),
 			HasValue:  true,
 			Interface: nil,
@@ -129,7 +135,7 @@ mismatch: [scalar]
 
 	t.Run("scalar", func(t *testing.T) {
 		var s string
-		run(TestCase{
+		run(t, TestCase{
 			Value:     provider.Get("practical.toyota"),
 			HasValue:  true,
 			Interface: "camry",
@@ -140,7 +146,7 @@ mismatch: [scalar]
 
 	t.Run("map", func(t *testing.T) {
 		var m map[string]string
-		run(TestCase{
+		run(t, TestCase{
 			Value:    provider.Get("practical"),
 			HasValue: true,
 			Interface: map[interface{}]interface{}{
@@ -159,7 +165,7 @@ mismatch: [scalar]
 
 	t.Run("slice", func(t *testing.T) {
 		var s []string
-		run(TestCase{
+		run(t, TestCase{
 			Value:     provider.Get("fun"),
 			HasValue:  true,
 			Interface: []interface{}{"maserati", "lamborghini"},
@@ -175,7 +181,7 @@ mismatch: [scalar]
 			Backseat  []string
 		}
 		var o Occupants
-		run(TestCase{
+		run(t, TestCase{
 			Value:    provider.Get("occupants.honda"),
 			HasValue: true,
 			Interface: map[interface{}]interface{}{
@@ -193,10 +199,12 @@ mismatch: [scalar]
 	})
 
 	t.Run("nil merged into scalar", func(t *testing.T) {
+		// Merging replaces scalars, so an explicit nil should replace
+		// lower-priority values.
 		t.Skip("TODO: ignores explicit nil")
 		var s string
-		run(TestCase{
-			Value:     provider.Get("antique"),
+		run(t, TestCase{
+			Value:     provider.Get("antique_scalar"),
 			HasValue:  true,
 			Interface: nil,
 			Populate:  &s,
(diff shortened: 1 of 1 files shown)


config_ext_test.go
```diff
@@ -0,0 +1,350 @@
...
+			Value:     val,
+			HasValue:  true,
+			Interface: "bottom",
+			Populate:  &s,
+		})
+		assert.Equal(t, "bottom", s, "unexpected result after populate")
+	})
+
+	t.Run("deep copy", func(t *testing.T) {
+		t.Skip("TODO: mutations are visible to other callers")
```
akshayjshah: This is probably the most serious of the existing bugs.

config_ext_test.go
```diff
@@ -0,0 +1,350 @@
...
+		run(TestCase{
+			Value:     provider.Get("occupants").Get("honda").Get("driver"),
+			HasValue:  true,
+			Interface: "jane",
+			Populate:  &s,
+		})
+		assert.Equal(t, "jane", s, "unexpected result after populate")
+	})
+
+	t.Run("default missing value", func(t *testing.T) {
```
amckinney: Is this test identical to the above?

amckinney: Based on the test below, I have a feeling this is supposed to test a call to `WithDefault` where the result would have otherwise been `nil`.

config_ext_test.go
```diff
@@ -0,0 +1,350 @@
...
+		})
+		assert.Equal(t, map[string]string{
+			"honda":  "civic",
+			"toyota": "camry",
+			"nissan": "altima",
+			"ford":   "fiesta",
+		}, m, "unexpected result after populate")
+	})
+
+	t.Run("default replaces sequences", func(t *testing.T) {
```
amckinney: The test description is a little misleading - the default doesn't replace the configured sequence in this case.

config_ext_test.go
```diff
@@ -0,0 +1,350 @@
...
+	t.Run("default ignores type mismatches", func(t *testing.T) {
+		// TODO: We should return an error in strict mode.
+		// fun is a sequence, so we shouldn't be able to merge in a mapping
+		val, err := provider.Get("fun").WithDefault(map[string]string{"dodge": "viper"})
+		assert.NoError(t, err, "error on type mismatch in permissive mode")
+		assert.Equal(t, []interface{}{"maserati", "lamborghini"}, val.Value(), "wrong merged output")
+	})
+
+	t.Run("chained defaults", func(t *testing.T) {
+		// Defaults should be FIFO: once we've set a default value, we can't later
```
amckinney: nit: not sure if FIFO is the best description here since that might imply the next-queued default would be used on a subsequent call to `Get` (i.e. `boom` would be used after `bottom`). Seems strange to use "idempotent" in this context, but something similar to "no effect after the first call", might be more appropriate.

config_ext_test.go
amckinney: ...oh no

amckinney:
Some minor comments, but otherwise looks great!

config_ext_test.go
akshayjshah: :man_facepalming: you're correct.

config_ext_test.go
akshayjshah: Good point, I'll change - how does "default sequences aren't merged" sound?

config_ext_test.go
akshayjshah: Also a good point, I'll reword this.

akshayjshah:
@SyrieBianco this is the config package we were talking about.

config_ext_test.go
```diff
@@ -0,0 +1,350 @@
...
+	})
+
+	t.Run("ignored field", func(t *testing.T) {
+		type cfg struct {
+			Boring []string `yaml:"-"`
+		}
+
+		var c cfg
+		require.NoError(t, provider.Get(Root).Populate(&c))
+		assert.Equal(t, 0, len(c.Boring), "wrong number of fun cars")
```
prashantv: optionally: `assert.Empty`

config_ext_test.go
```diff
@@ -0,0 +1,350 @@
...
+  - porsche
+practical:
+  toyota: camry
+  honda: accord
+occupants:
+  honda:
+    driver: jane
+    backseat: [nate]
+`)
+	second := strings.NewReader(`
```
prashantv: can we also add a test where a value in `first` is set to `~` in `second`, ensuring that overrides to `nil` work

config_ext_test.go
```diff
@@ -0,0 +1,350 @@
...
+			HasValue:  true,
+			Interface: nil,
+			Populate:  &s,
+		})
+		assert.Equal(t, "", s, "unexpected result after populate")
+	})
+
+	t.Run("default merges maps", func(t *testing.T) {
+		var m map[string]string
+		val, err := provider.Get("practical").WithDefault(map[string]string{"ford": "fiesta"})
```
prashantv: would also add a key that exists in `first` to make sure that the merge replaces existing keys from defaults

config_ext_test.go
```diff
@@ -0,0 +1,350 @@
...
+			"nissan": "altima",
+		}, m, "unexpected result after populate")
+	})
+
+	t.Run("slice", func(t *testing.T) {
+		var s []string
+		run(TestCase{
+			Value:     provider.Get("fun"),
+			HasValue:  true,
+			Interface: []interface{}{"maserati", "lamborghini"},
```
prashantv: for some reason, i thought that the custom merging logic merged lists, rather than replacing, but i guess it matches the yaml unmarshaler?

config_ext_test.go
```diff
@@ -0,0 +1,350 @@
...
+	"testing"
+
+	. "go.uber.org/config"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestIntegration(t *testing.T) {
+	first := strings.NewReader(`
```
prashantv: is it worth using the yaml pointers to make sure they work as expected?

config_ext_test.go
amckinney: 👍

config_ext_test.go
akshayjshah: For no particular reason, I usually avoid that helper. Alex mentioned the same thing, so I'll change.

config_ext_test.go
akshayjshah: Good point, definitely worthwhile.

config_ext_test.go
akshayjshah: Yep. There are a bunch of internal tickets where we discussed merging lists, but the actual implementation appears to match the yaml package pretty closely.

config_ext_test.go
akshayjshah: Done.

config_ext_test.go
akshayjshah: *Great* call. As it turns out, that test fails...so the current behavior of `WithDefault` is to merge in new keys, but not replace existing keys. I'm calling this a bug rather than a behavior to preserve.

akshayjshah:
Pushed another commit addressing CR feedback. I've included tests for errors later in the stack...but they're definitely coming.

config_ext_test.go
```diff
@@ -31,6 +35,7 @@ fun:
 practical:
   honda: civic
   nissan: altima
+antique: model_t
```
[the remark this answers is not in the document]
akshayjshah: #72 is covered explicitly below, applying a non-nil default to an explicitly nulled-out value. Can easily switch this around too.

config_ext_test.go
```diff
@@ -0,0 +1,426 @@
...
+		})
+		assert.Equal(t, Occupants{
+			Driver:    "jane",
+			Passenger: "arthur",
+			Backseat:  []string{"nora"},
+		}, o, "unexpected result after populate")
+	})
+
+	t.Run("nil merged into scalar", func(t *testing.T) {
+		t.Skip("TODO: ignores explicit nil")
```
prashantv: thanks for adding this -- one other case that might be worth adding is nil merging into an existing sequence/map. I'm not sure of the right behaviour, I'd assume replace in both cases, but since maps merge, perhaps it just gets ignored?

akshayjshah: Good point. I'd expect sequences to be replaced and maps to be merged, since that's the behavior for non-nil things.

config_ext_test.go
```diff
@@ -0,0 +1,461 @@
...
+		// lower-priority values unchanged.
+		var m map[string]string
+		run(t, TestCase{
+			Value:     provider.Get("antique_mapping"),
+			HasValue:  true,
+			Interface: map[interface{}]interface{}{"ford": "model_t"},
+			Populate:  &m,
+		})
+		assert.Equal(t, map[string]string{"ford": "model_t"}, m, "unexpected result after Populate")
+	})
```
akshayjshah: In Go, this is consistent with the rest of the merging behavior - maps are deep-merged, and `nil` is a valid map with no keys.

However, it's a little weird. I'm open to the idea that we treat nils and empty maps differently here - `nil` replaces the existing value, and `{}` has no effect.

prashantv: I'd prefer that behaviour, but we should probably look at what the YAML library does in this case. We may want to match it (since the rest of the library seems to try and match it)?

config_ext_test.go
```diff
@@ -0,0 +1,461 @@
...
+		assert.Equal(t, []string{"maserati", "lamborghini"}, s, "unexpected result after populate")
+	})
+
+	t.Run("default ignores type mismatches", func(t *testing.T) {
+		// TODO: We should return an error in strict mode.
+		// fun is a sequence, so we shouldn't be able to merge in a mapping
+		val, err := provider.Get("fun").WithDefault(map[string]string{"dodge": "viper"})
+		assert.NoError(t, err, "error on type mismatch in permissive mode")
+		assert.Equal(t, []interface{}{"maserati", "lamborghini"}, val.Value(), "wrong merged output")
+	})
```
akshayjshah: This is wrong; defaults should act like a `base_base.yaml`, which controls the type of each key. The end result here should be `{dodge: viper}`.

config_ext_test.go
```diff
@@ -0,0 +1,461 @@
...
+
+	t.Run("chained defaults", func(t *testing.T) {
+		// Once we've set a default, the key has a value. Subsequent calls to
+		// WithDefault have no effect.
+		val, err := provider.Get("top").WithDefault(map[string]string{"middle": "bottom"})
+		require.NoError(t, err, "couldn't set first default")
+		val, err = val.Get("middle").WithDefault("boom")
```
akshayjshah: Should be same type as first `WithDefault`.

For consistency, we should also be treating *each* `WithDefault` as though it were the base for a merge. Here, we should be able to set `{other_middle: other_bottom}` in a second call to `WithDefault` and see both keys and values in the final result.

config_ext_test.go
akshayjshah: Yeah - after talking through this with Alex, we came to the same conclusion. That's also how the upstream YAML lib behaves, so we're safe.

PR #81, opened 2018-05-31
Add integration tests for NewValue

akshayjshah:
*WIP, but please take a look and discuss.*

This is...unfortunate. Unless users are very careful, the behavior of
`NewValue` is completely irrational. These tests aren't exhaustive, but
the gist is this:

* `NewValue` takes a provider, a key, a value (as `interface{}`), and a
bool indicating whether the value was found or not. There's currently no
validation checking that `value` and `found` match the contents of the
provider at the specified key.
* `HasValue` returns the user-supplied bool, ignoring the contents of
the provider.
* `String` and `Value` return the user-supplied value, ignoring the
contents of the provider.
* `Populate`, `Get`, and `WithDefault` use the provider, ignoring the
user-supplied value and bool.

This leads to a lot of potentially counter-intuitive behavior. Since the
primary use case for `NewValue` is in tests, this is misleading - the
test may not reflect reality at all.

I'm opening this PR to let everyone ponder a good path forwards. There
are 90ish uses of `NewValue` internally, but very few of them use the
`Value` or `HasValue` APIs. I propose this: in `NewValue`, validate that
`value` reflects the contents of the provider at the specified key. If
it doesn't, panic (I know, I know). Ignore `found` completely. Once
validation is complete, just return `provider.Get(key)`. Before cutting
a new release, deprecate `NewValue` and many of the methods on `Value`,
including `Value()` and `HasValue()`.

This keeps most of the API compatible, makes the behavior sane-ish, and
isn't likely to break anyone's production code. Thoughts?

### DIFF
--- a/value_ext_test.go
+++ b/value_ext_test.go
@@ -64,7 +64,7 @@ func TestValueIntegration(t *testing.T) {
 			true,    // found
 		)
 		assert.Equal(t, "hello", v.Value(), "expected to use user-supplied value")
-		assert.Equal(t, "foo", v.Get("scalar").Value(), "get exposes data not in top-level value") // WAT?!
+		assert.Equal(t, "foo", v.Get("scalar").Value(), "get exposes data not in top-level value") // WAT
 	})
 
 	t.Run("found overrides provider", func(t *testing.T) {
@@ -74,17 +74,21 @@ func TestValueIntegration(t *testing.T) {
 			"foo",    // value
 			false,    // found, doesn't match provider
 		)
-		assert.False(t, v.HasValue(), "unexpected string representation")
+		assert.False(t, v.HasValue(), "HasValue should be false")
+		assert.Equal(t, "foo", v.Value(), "Value should return something") // WAT
 	})
 
-	t.Run("found overrides provider", func(t *testing.T) {
+	t.Run("default re-exposes provider", func(t *testing.T) {
 		v := NewValue(
 			provider,
 			"scalar", // key
-			"foo",    // value
-			false,    // found, doesn't match provider
+			"hello",  // value, doesn't match provider contents
+			true,     // found
 		)
-		assert.False(t, v.HasValue(), "unexpected string representation")
+		assert.Equal(t, "hello", v.Value(), "expected to use user-supplied value")
+		defaulted, err := v.WithDefault("goodbye")
+		require.NoError(t, err, "couldn't set default")
+		assert.Equal(t, "foo", defaulted.Value(), "expected to re-expose provider") // WAT
 	})
 
 	t.Run("populate", func(t *testing.T) {
@@ -95,7 +99,8 @@ func TestValueIntegration(t *testing.T) {
 			"bar",    // value, doesn't match provider contents
 			true,     // found
 		)
+		assert.Equal(t, "bar", v.Value(), "expected to use user-supplied value")
 		require.NoError(t, v.Populate(&s), "error on first populate")
-		assert.Equal(t, "foo", s, "expected to use provider-contained value")
+		assert.Equal(t, "foo", s, "expected to use provider-contained value") // WAT
 	})
 }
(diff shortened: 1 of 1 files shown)


value_ext_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+	t.Run("get re-exposes provider", func(t *testing.T) {
+		v := NewValue(
+			provider,
+			Root,    // key
+			"hello", // value, doesn't match provider contents
+			true,    // found
+		)
+		assert.Equal(t, "hello", v.Value(), "expected to use user-supplied value")
+		assert.Equal(t, "foo", v.Get("scalar").Value(), "get exposes data not in top-level value") // WAT?!
```
prashantv: wat indeed.. i really don't understand why a value can't just use the underlying source to always get the value.

value_ext_test.go
```diff
@@ -0,0 +1,101 @@
...
+	})
+
+	t.Run("found overrides provider", func(t *testing.T) {
+		v := NewValue(
+			provider,
+			"scalar", // key
+			"foo",    // value
+			false,    // found, doesn't match provider
+		)
+		assert.False(t, v.HasValue(), "unexpected string representation")
```
prashantv: if you call `.Value()` or `.Get(..).Value()`, is there values?

value_ext_test.go
```diff
@@ -0,0 +1,101 @@
...
+	})
+
+	t.Run("found overrides provider", func(t *testing.T) {
```
prashantv: think this is the exact same test as above?

value_ext_test.go
```diff
@@ -0,0 +1,101 @@
...
+		)
+		assert.False(t, v.HasValue(), "unexpected string representation")
+	})
+
+	t.Run("populate", func(t *testing.T) {
+		var s string
+		v := NewValue(
+			provider,
+			"scalar", // key
+			"bar",    // value, doesn't match provider contents
```
prashantv: so the value passed in is arbitrarily ignored based on which API is used to access the value?

value_ext_test.go
akshayjshah: It definitely could. From what I can infer, keeping the value at the current key as a struct field was just an optimization - the current provider implementation takes care to keep the two in sync. This constructor would have been fine as an unexported helper, but we didn't catch quite how unpredictable the behavior is in the pre-1.0 review.

value_ext_test.go
akshayjshah: Of course there are! I'll update the diff to point that out.

value_ext_test.go
akshayjshah: Yep...I think I was partway through adding some more tests when I ran out yesterday.

value_ext_test.go
akshayjshah: Indeed. "Arbitrary" is exactly the right word here, too - there's no rhyme or reason to it.

value_ext_test.go
akshayjshah: Fixed, replaced this to demonstrate some more WAT in the behavior of `WithDefault`.

value_ext_test.go
```diff
@@ -0,0 +1,106 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestValueIntegration(t *testing.T) {
+	provider, err := NewStaticProvider(map[string]interface{}{
+		"scalar":   "foo",
+		"mapping":  map[string]int{"one": 1, "two": 2},
+		"sequence": []int{1, 2},
```
akshayjshah: Should remove `mapping` and `sequence`, since they're not required to demonstrate all the problems below.

value_ext_test.go
```diff
@@ -0,0 +1,106 @@
...
+package config
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestValueIntegration(t *testing.T) {
```
akshayjshah: Global comment: restructure so tests are grouped by inconsistency (e.g., `found mismatch`) rather than the API being tested.

PR #82, opened 2018-06-05
Small Makefile cleanups

akshayjshah:
Clean up a few more odds and ends in the Makefile, including a
mistakenly commented-out line.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-PACKAGES := $(shell glide novendor)
+SHELL := /usr/bin/env bash
 TEST := go test -race
 
 .PHONY: install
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -8,14 +8,13 @@ install:
...
+	$(TEST) ./...
 
 .PHONY: ci
-ci: SHELL := /bin/bash
 ci:
 ifdef COVER
-	$(TEST) -cover -coverprofile=coverage.txt $(PACKAGES)
-	# bash <(curl -s https://codecov.io/bash)
+	$(TEST) -cover -coverprofile=coverage.txt ./...
+	bash <(curl -s https://codecov.io/bash)
```
amckinney: Should this be `$(SHELL) <(curl -s https://codecov.io/bash)` ?

akshayjshah: No, the `SHELL` variable is special: https://www.gnu.org/software/make/manual/html_node/Choosing-the-Shell.html

PR #83, opened 2018-06-05
Clean up integration tests

akshayjshah:
Pull the common portion of the integration tests out into a separate
function, which makes it easier to run the same set of tests on multiple
providers. Then, add separate test suites for the current permissive
provider logic, a yet-to-be-added strict provider, and a static
provider.

Along the way, fix some logical bugs in the tests:

* In one test, the assertion is just wrong - the user-supplied config
  should be overwriting data supplied via `WithDefault`.
* In another, we're mistakenly drilling down with `Value.Get` but making
  assertions about the top-level data.
* In a third, we're making incorrect assertions about the output of a
  deep merge. The current provider is correct, so we don't need to skip
  the test.
* Most importantly, we're needlessly changing the behavior of the YAML
  provider when it encounters a type mismatch between lower- and
  higher-priority sources. No meaningful merge is possible, and today, the
  YAML provider overwrites the lower-priority data with the higher. This
  is reasonable, so we should preserve that behavior.

### DIFF
--- a/config_ext_test.go
+++ b/config_ext_test.go
@@ -28,10 +28,11 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	yaml "gopkg.in/yaml.v2"
 )
 
-func TestIntegration(t *testing.T) {
-	first := strings.NewReader(`
+const (
+	_base = `
 nothing: ~
 fun:
   - maserati
@@ -53,9 +54,8 @@ occupants:
 extra_practical:
   <<: *ptr
   volkswagon: jetta
-mismatch: scalar
-`)
-	second := strings.NewReader(`
+`
+	_override = `
 fun:
   - maserati
   - lamborghini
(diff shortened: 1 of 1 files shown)


config_ext_test.go
```diff
@@ -447,57 +418,210 @@ mismatch: [scalar]
...
+		strings.NewReader(_override),
+	)
+	require.NoError(t, err, "couldn't create provider")
+	assert.Equal(t, `cached "yaml"`, provider.Get(Root).Source(), "wrong source")
+	assert.Equal(t, "<nil>", provider.Get("nothing").String(), "wrong string representation")
 
-		var c cfg
-		require.NoError(t, provider.Get(Root).Populate(&c))
-		assert.Equal(t, 2, len(c.Fun), "wrong number of fun cars")
+	runCommonTests(t, provider)
```
amckinney: nice!

config_ext_test.go
```diff
@@ -447,57 +418,210 @@ mismatch: [scalar]
...
+		assert.NoError(t, err, "error on type mismatch in permissive mode")
+		assert.Equal(t, "foo", val.Value(), "wrong merged output")
+	})
 
-		var c cfg
-		require.NoError(t, provider.Get(Root).Populate(&c))
-		assert.Equal(t, 2, len(c.Fun), "wrong number of fun cars")
+	t.Run("ignore duplicate keys", func(t *testing.T) {
+		provider, err := NewYAMLProviderFromReader(
+			strings.NewReader("dupe: foo\ndupe: bar"),
```
amckinney: The last configured wins?

config_ext_test.go
```diff
@@ -447,57 +418,210 @@ mismatch: [scalar]
...
+
+func TestStaticFromYAML(t *testing.T) {
+	// Since we have a common test suite, we may as well use it to exercise the
+	// static provider too.
+	var base, override interface{}
+	require.NoError(t, yaml.Unmarshal([]byte(_base), &base), "couldn't unmarshal base YAML")
+	require.NoError(t, yaml.Unmarshal([]byte(_override), &override), "couldn't unmarshal base YAML")
+
+	t.Skip("TODO: no support for multiple sources in NewStaticProvider")
+	p := NopProvider{}
```
amckinney: Will this change to `NewStaticProvider` once we fix the test?

config_ext_test.go
akshayjshah: Yep. That's what the YAML specification calls for :man_shrugging:

config_ext_test.go
akshayjshah: Sort of - it'll change to a `NewYAML` with two different sources of static configuration.

PR #85, opened 2018-06-05
Add a small package to deep-merge YAML

akshayjshah:
Since we're only planning to use this package during process startup,
there's no need to micro-optimize our YAML merging. Given that, we can
implement our deep-merging logic in a small, standalone package that
doesn't require any explicit reflection. That package takes a more naive
approach than our current logic: it unmarshals all YAML sources into
`interface{}`, merges the resulting Go values, and re-serializes the
merged `interface{}` to YAML.

### DIFF
--- a/internal/merge/merge.go
+++ b/internal/merge/merge.go
@@ -49,10 +49,9 @@ type (
 //   == {"foo": [4, 5, 6]}
 //
 // In non-strict mode, duplicate map keys are allowed within a single source,
-// with later values overwriting previous ones. Similarly, attempting to merge
-// mismatched types (e.g., merging a sequence into a map) silently fails. To
-// preserve backward compatibility during type mismatches, new values replace
-// existing values.
+// with later values overwriting previous ones. Attempting to merge
+// mismatched types (e.g., merging a sequence into a map) replaces the old
+// value with the new.
 //
 // Enabling strict mode returns errors in both of the above cases.
 func YAML(sources []io.Reader, strict bool) (*bytes.Buffer, error) {
@@ -73,7 +72,7 @@ func YAML(sources []io.Reader, strict bool) (*bytes.Buffer, error) {
 		merged = pair
 	}
 
-	buf := bytes.NewBuffer(nil)
+	buf := &bytes.Buffer{}
 	enc := yaml.NewEncoder(buf)
 	if err := enc.Encode(merged); err != nil {
 		return nil, unreachable.Wrap(fmt.Errorf("couldn't re-serialize merged YAML: %v", err))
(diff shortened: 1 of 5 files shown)


internal/merge/merge.go
```diff
@@ -0,0 +1,151 @@
...
+// Maps are deep-merged. For example,
+//   {"one": 1, "two": 2} + {"one": 42, "three": 3}
+//   == {"one": 42, "two": 2, "three": 3}
+// Sequences are replaced. For example,
+//   {"foo": [1, 2, 3]} + {"foo": [4, 5, 6]}
+//   == {"foo": [4, 5, 6]}
+//
+// In non-strict mode, duplicate map keys are allowed within a single source,
+// with later values overwriting previous ones. Similarly, attempting to merge
+// mismatched types (e.g., merging a sequence into a map) silently fails. To
```
amckinney: Not sure if "silently fails" is the right wording here, seeing as the behavior is described below. Maybe just rephrase to:
```
Similarly, in order to preserve backwards compatibility, new values will replace existing values when mismatched types are merged.
```

internal/merge/merge.go
```diff
@@ -0,0 +1,151 @@
...
+
+	buf := bytes.NewBuffer(nil)
+	enc := yaml.NewEncoder(buf)
+	if err := enc.Encode(merged); err != nil {
+		return nil, unreachable.Wrap(fmt.Errorf("couldn't re-serialize merged YAML: %v", err))
+	}
+	return buf, nil
+}
+
+func merge(left, right interface{}, strict bool) (interface{}, error) {
```
amckinney: For clarity, I've seen this written as `mergeRight(...)` before. We might want to consider adding that here.

amckinney: Alternatively, we could rename the parameters to `merge(into, from interface{}, strict bool)`

internal/merge/merge.go
```diff
@@ -0,0 +1,151 @@
...
+		}
+
+		pair, err := merge(merged, contents, strict)
+		if err != nil {
+			return nil, err // error is already descriptive enough
+		}
+		merged = pair
+	}
+
+	buf := bytes.NewBuffer(nil)
```
amckinney: nit: From [the docs](https://golang.org/pkg/bytes/#NewBuffer), we might be able to simply use `var buf *bytes.Buffer` here.

internal/merge/merge.go
```diff
@@ -0,0 +1,151 @@
...
+	if isMapping(left) && isMapping(right) {
+		return mergeMapping(left.(mapping), right.(mapping), strict)
+	}
+	// YAML types don't match, so no merge is possible. For backward
+	// compatibility, ignore mismatches unless we're in strict mode and return
+	// the higher-priority value.
+	if !strict {
+		return right, nil
+	}
+	return nil, fmt.Errorf("can't merge a %s and a %s", describe(left), describe(right))
```
amckinney: Let's add some clarity in this error message:
```
fmt.Errorf("can't merge a %s into a %s", describe(right), describe(left))
```

internal/merge/merge_test.go
```diff
@@ -0,0 +1,212 @@
...
+		})
+		t.Run(tt.desc+" permissive", func(t *testing.T) {
+			// prefer the higher-priority value
+			succeeds(t, false, tt.left, tt.right, tt.right)
+		})
+	}
+}
+
+func TestBooleans(t *testing.T) {
+	// YAML helpfully interprets many strings as Booleans.
```
amckinney: Hmm? The more you know!

internal/merge/testdata/expect.yaml
```diff
@@ -0,0 +1,13 @@
+fun:
+- lamborghini
+- porsche
+occupants:
+  honda:
+    backseat:
+    - nora
```
amckinney: nit: indent here

internal/merge/merge.go
akshayjshah: Sounds good.

internal/merge/merge.go
akshayjshah: Sounds good, changed.

internal/merge/merge.go
akshayjshah: Both are common; for whatever reason, I habitually use the constructor. Happy to change if you'd prefer.

internal/merge/merge.go
akshayjshah: Good idea.

internal/merge/merge_test.go
akshayjshah: YAML, you so crazy.

akshayjshah:
Addressed feedback in https://github.com/uber-go/config/pull/85/commits/3a3f3f5919591d6157d1e41e21456652def18729.

internal/merge/merge.go
```diff
@@ -73,51 +72,51 @@ func YAML(sources []io.Reader, strict bool) (*bytes.Buffer, error) {
...
 	// to merge whole YAML files. Since we're always unmarshaling into
 	// interface{}, we only need to handle a few types. This ends up being
 	// cleaner if we just handle each case explicitly.
-	if left == nil {
-		return right, nil
+	if into == nil {
+		return from, nil
 	}
-	if right == nil {
+	if from == nil {
```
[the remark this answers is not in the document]
akshayjshah: I think I like it as-is - each branch is closely tied to one of the integration tests. I hadn't considered this, though - nice catch!

Issue #57, opened 2017-08-07
Report merge error during provider group construction


ghost:
Provider group should report merge errors on construction, not when trying to retrieve a value via `Get` method, because we need to define invalid Values then and show how users can work with them.

The difficulty here is to flatten maps, e.g. if we have `base.yaml`:
```yaml
series:
  episodes:
    s01e01: Pilot
```
and `test.yaml`:
```yaml
series.episodes:
- Pilot
```
The provider should fail to merge these files together because `base.yaml` has `series.episodes` key defined as a map and `test.yaml` defines it as a slice. This test should pass:

```golang
func TestProviderGroup_ConstructorMergeError(t *testing.T) {
	t.Parallel()

	f, err := NewStaticProvider(map[string]interface{}{
		"series": map[string]interface{}{
			"episodes": map[string]string{"s01e01": "Pilot"},
		}})

	require.NoError(t, err, "Can't create the first provider")

	s, err := NewStaticProvider(map[string]interface{}{
		"series.episodes": []interface{}{"Pilot"},
	})
	require.NoError(t, err, "Can't create the second provider")

	_, err := NewProviderGroup("test-group", f, s)
	assert.Error(t, err)
}
```

akshayjshah:
Fixed in v1.2.0.

Issue #80, opened 2018-05-31
Environment variable expansion also applies to comments


akshayjshah:
We attempt to expand environment variable references in comments, which is irritating - constructing a provider fails if any of the referenced variables aren't available. For example, this YAML can't be loaded because `$ZONE` isn't available:

```yaml
# We use $DATACENTER because the more-correct $ZONE isn't available yet.
tag: ${DATACENTER:unknown}
```

akshayjshah:
Fixed in v1.2.0.

Issue #73, opened 2017-12-15
NopProvider returns true for `HasValue()`


glibsm:
The following test fails, yet it should succeed since NopProvider by definition can not have values.

```
package config

import (
  "testing"

  "github.com/stretchr/testify/require"
)

func TestNopProviderHasValue(t *testing.T) {
  p := NopProvider{}
  val := p.Get("some.obviously.missing.value")
  require.False(t, val.HasValue(), "nop provider can't have values")
}
```

glibsm:
There is an existing test in the codebase that actually explicitly tests for `HasValue()` to be `true`:

```go
func TestNopProvider_Get(t *testing.T) {
	t.Parallel()

	p := NopProvider{}
	assert.Equal(t, "NopProvider", p.Name())

	v := p.Get("randomKey")
	assert.Equal(t, "NopProvider", v.Source())
	assert.True(t, v.HasValue())
	assert.Nil(t, v.Value())
}
```

This is very surprising behavior to me. Is that intentional?

akshayjshah:
Fixed in v1.2.0.

PR #87, opened 2018-06-05
Replace provider and value implementations

akshayjshah:
With the foundation laid by an integration test suite and the internal
YAML-merging package, gut the internals of this package and reimplement
the core types.

The new implementation is intentionally naive: since we're only using
this package at process startup, we don't care much about performance
and can afford repeated round-trips through YAML. This has two benefits:
the code is simpler, and it's easier to leverage all the features that
`gopkg.in/yaml.v2` already has (especially strict unmarshalling mode).
While we have the opportunity, we also introduce a new constructor that
takes functional options. I've also improved the GoDoc and deprecated a
bunch of the existing APIs (though they all continue to function as
expected).

This fixes issues #57, #59, #72, #73, #75, #76, #80, and a pile of other
bugs that weren't reported by users.


akshayjshah:
screenshot_2018-06-05 replace provider and value implementations by akshayjshah pull request 87 uber-go config

config.go
```diff
@@ -0,0 +1,326 @@
...
+	same, err := areSameYAML(contents, value)
+	if err != nil {
+		panic(fmt.Sprintf("can't check NewValue parameter consistency: %v", err))
+	}
+	if !same {
+		tmpl := "inconsistent parameters: provider %s has %#v at key %q but value was %#v"
+		panic(fmt.Sprintf(tmpl, p.Name(), contents, key, value))
+	}
+	return p.Get(key)
+}
```
akshayjshah: @prashantv and @amckinney: as proposed in #81, here's the new implementation of `NewValue`. Behavior of the returned value is now always the same as `Provider.Get`, at the cost of a panic when the user-supplied parameters are internally inconsistent.

constructors.go
```diff
@@ -0,0 +1,167 @@
...
+var _ Provider = (*scopedProvider)(nil)
+
+func (s *scopedProvider) Get(key string) Value {
+	return s.Provider.Get(fmt.Sprintf("%s.%s", s.prefix, key))
+}
+
+// NewScopedProvider wraps a provider and adds a prefix to all Get calls.
+func NewScopedProvider(prefix string, provider Provider) Provider {
+	if prefix == "" {
+		return provider
```
amckinney: Should we create a new provider anyways? The signature implies that we are creating a new instance so this would remove any possibility of mutation (e.g. a custom `config.Provider` implementation from the user).

```
func NewScopedProvider(prefix string, provider Provider) Provider {
    return &scopedProvider{provider, prefix /* can be "" */}
}
```

constructors.go
```diff
@@ -0,0 +1,167 @@
...
+// overriding earlier ones. The merge logic is described in the package-level
+// documentation. To preserve backward compatibility, the resulting provider
+// disables strict unmarshalling.
+//
+// Deprecated: compose multiple sources of configuration using NewYAML and the
+// Static, File, and/or Source options directly. This enables strict
+// unmarshalling by default and allows use of other options at the same time.
+func NewProviderGroup(name string, providers ...Provider) (Provider, error) {
+	opts := make([]YAMLOption, 0, len(providers)+2)
+	opts = append(opts, Name(name), Permissive())
```
amckinney: Ah, I see the use of the `Name` option now. Feel free to ignore my other comment.

constructors.go
```diff
@@ -0,0 +1,167 @@
...
+//
+// Deprecated: use NewYAML and the Static and Expand options directly. This
+// enables strict unmarshalling by default and allows use of other options at
+// the same time.
+func NewStaticProviderWithExpand(data interface{}, lookup LookupFunc) (Provider, error) {
+	return NewYAML(Static(data), Expand(lookup), Permissive())
+}
+
+// NewYAMLProviderFromBytes merges multiple YAML-formatted byte slices into a
+// single provider. As always, later configuration blobs override earlier ones
```
amckinney: nit: Consider dropping `As always` since the other functions don't include it and this might be the first function the reader is viewing.

equal.go
```diff
@@ -18,58 +18,25 @@
...
-  ports:
-  - 443
-`))
-
+// areSameYAML checks whether two values represent the same YAML data.
+func areSameYAML(left, right interface{}) (bool, error) {
+	l, err := yaml.Marshal(left)
 	if err != nil {
-		log.Fatal(err)
+		// unreachable with YAML provider, but possible with third-party provider
```
amckinney: How is a third-party provider relevant here? Aren't we just calling `gopkg.in/yaml.v2.Marshal` on a configuration type without a provider?

config.go
```diff
@@ -0,0 +1,329 @@
...
+		}
+		if _, ok := m[segment]; !ok {
+			return nil, false
+		}
+		cur = m[segment]
+	}
+	return cur, true
+}
+
+func (y *YAML) populate(path []string, i interface{}) error {
```
amckinney: For strict mode: Do we check if all the keys in `interface{}` are populated here?

example_test.go
```diff
@@ -0,0 +1,98 @@
...
+		"key": "value",
+	}))
+	if err != nil {
+		panic(err)
+	}
+	// Using config.Root as a key retrieves the whole configuration.
+	base := provider.Get(config.Root)
+
+	// Applying a default is equivalent to serializing it to YAML, writing the
+	// serialized bytes to default.yaml, and then merging default.yaml and the
```
amckinney: nit: and then merging the existing configuration into the `default.yaml`

expand.go
```diff
@@ -22,16 +22,64 @@ package config
...
+			key = in[:sep]
+			def = in[sep+1:]
+		}
+
+		if envVal, ok := lookUp(key); ok {
+			return envVal, nil
+		}
+
+		if def == "" {
+			return "", fmt.Errorf(`default is empty for %q (use "" for empty string)`, key)
```
amckinney: What if the key isn't found and it is of form `{$KEY}`? In this case, `def` will be `""`, but we will say `"default is empty ..."`. The default might be intentionally empty. Would it be better to say `$KEY could not be determined from the environment, or something similar?

option.go
```diff
@@ -0,0 +1,144 @@
...
+// all keys present in the YAML are allowed. Finally, type conflicts are
+// allowed when merging source files, with later values replacing earlier
+// ones.
+func Permissive() YAMLOption {
+	return optionFunc(func(c *config) {
+		c.strict = false
+	})
+}
+
+// Name customizes the name of the provider. The default name is "YAML".
```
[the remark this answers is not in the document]
amckinney: Ignore this one (commented elsewhere).

config.go
```diff
@@ -0,0 +1,329 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
```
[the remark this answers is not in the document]
akshayjshah: Guidance from legal (which matches most other corporate OSS) is that each file's header should include all the years in which the file was modified. Successive years can be coalesced into a range.

E.g.,
* 2017
* 2015, 2017-2018
* 2015-2018

config.go
```diff
@@ -0,0 +1,329 @@
...
+	}
+
+	if cfg.err != nil {
+		return nil, fmt.Errorf("error applying options: %v", cfg.err)
+	}
+
+	// On construction, go through a full merge-serialize-deserialize cycle to
+	// catch any duplicated keys as early as possible (in strict mode). It also
+	// strips comments, which stops us from attempting environment variable
+	// expansion.
```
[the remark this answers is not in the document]
akshayjshah: Good point, will do.

config.go
```diff
@@ -0,0 +1,329 @@
...
+
+	if cfg.lookup != nil {
+		exp, err := ioutil.ReadAll(transform.NewReader(merged, newExpandTransformer(cfg.lookup)))
+		if err != nil {
+			return nil, fmt.Errorf("couldn't expand environment: %v", err)
+		}
+		merged = bytes.NewBuffer(exp)
+	}
+
+	raw := merged.Bytes() // helpful for WithDefault later on
```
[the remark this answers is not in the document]
akshayjshah: Unfortunately, we can't do that because we're using the buffer as a reader a few lines down. Once all the contents of the buffer are consumed via that API, `.Bytes()` returns an empty slice.

We could capture the slice here and use `Unmarshal`/`UnmarshalStrict` below, but I thought that was a little uglier.

config.go
```diff
@@ -0,0 +1,329 @@
...
+//   baz: hello
+//
+// To get a value holding the entire configuration, use the Root constant as
+// the key.
+func (y *YAML) Get(key string) Value {
+	return y.get(strings.Split(key, _separator))
+}
+
+func (y *YAML) get(path []string) Value {
+	if len(path) == 1 && path[0] == "" {
```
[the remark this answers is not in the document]
akshayjshah: Definitely.

config.go
```diff
@@ -0,0 +1,329 @@
...
+		path = nil // root node
+	}
+	return Value{
+		path:     path,
+		provider: y,
+	}
+}
+
+// at returns the unmarshalled representation of the value at a given path,
+// with a bool indicating whether the value was found.
```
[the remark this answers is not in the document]
akshayjshah: Done.

config.go
akshayjshah: No, that's built into the YAML library. When we call `SetStrict` on decoders, we enable these checks.

config.go
```diff
@@ -0,0 +1,329 @@
...
+	}
+	sources := []io.Reader{rawDefault, bytes.NewBuffer(y.raw)}
+	merged, err := merge.YAML(sources, y.strict)
+	if err != nil {
+		// The defaults and the existing config don't agree on the type of a value
+		// (e.g., we're defaulting a mapping to a sequence, which doesn't make
+		// sense).
+		return nil, fmt.Errorf("merging default and existing YAML failed: %v", err)
+	}
+	rawMerged := merged.Bytes()
```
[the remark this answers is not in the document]
akshayjshah: Same issue as above, unfortunately.

config.go
```diff
@@ -0,0 +1,329 @@
...
+	contents := actual.Value()
+	same, err := areSameYAML(contents, value)
+	if err != nil {
+		panic(fmt.Sprintf("can't check NewValue parameter consistency: %v", err))
+	}
+	if !same {
+		tmpl := "inconsistent parameters: provider %s has %#v at key %q but value was %#v"
+		panic(fmt.Sprintf(tmpl, p.Name(), contents, key, value))
+	}
+	return p.Get(key)
```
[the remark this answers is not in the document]
akshayjshah: Good call.

constructors.go
akshayjshah: I don't think that's necessary; if the provider implementation mutates its internal state when users call the exported API, we'll have problems anyways (since the `scopedProvider` is just delegating to the user-supplied provider).

constructors.go
akshayjshah: Good point, will do.

equal.go
akshayjshah: I'll clarify the docs here. We only call this function from `NewValue`, where the provider is user-supplied. Since we can't guarantee that the user-supplied provider is the one from this package, we can't absolutely guarantee that the value is representable in YAML.

option.go
```diff
@@ -0,0 +1,144 @@
...
+
+// Expand enables variable expansion in all provided sources. The supplied
+// function MUST behave like os.LookupEnv: it looks up a key and returns a
+// value and whether the key was found.
+//
+// Expand allows variable references to take two forms: $VAR or
+// ${VAR:default}. In the first form, variable names MUST adhere to shell
+// naming rules:
+//   ...a word consisting solely of underscores, digits, and alphabetics form
+//   the portable character set. The first character of a name is not a digit.
```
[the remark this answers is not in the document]
akshayjshah: :+1:

option.go
```diff
@@ -0,0 +1,144 @@
...
+// Permissive disables gopkg.in/yaml.v2's strict mode. It's provided for
+// backward compatibility; to avoid a variety of common mistakes, most users
+// should leave YAML providers in the default strict mode.
+//
+// In permissive mode, duplicate keys in the same source file are allowed.
+// Later values override earlier ones (note that duplicates are NOT merged,
+// unlike all other merges in this package). Calls to Populate that don't use
+// all keys present in the YAML are allowed. Finally, type conflicts are
+// allowed when merging source files, with later values replacing earlier
+// ones.
```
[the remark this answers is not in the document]
akshayjshah: :) I try.

expand.go
akshayjshah: Possibly. This whole function was copied verbatim from a now-deleted file - I'd prefer to keep it exactly as-is right now.

The `expand.go` file is a bit of a mess, and it'll need a round of refactoring just so the project passes lint. We can address this stuff during that pass.

config.go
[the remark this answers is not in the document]
amckinney: Cool, thanks for the clarification.

config.go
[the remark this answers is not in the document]
amckinney: Ah got it, I wasn't aware. Thanks for clarifying! I prefer this implementation over your suggestion.

constructors.go
amckinney: 👍

expand.go
amckinney: Sounds good!

PR #89, opened 2018-06-27
Fix deep map merging for missing keys in dst

dgladkov:
This change is based on 1.1.0 release. This fixes a case when map from `src` gets assigned to a key in `dst` and then mutated in subsequent `mergeMaps` call.

### DIFF
--- a/yaml.go
+++ b/yaml.go
@@ -91,16 +91,13 @@ func newYAMLProviderCore(files ...io.Reader) (*yamlConfigProvider, error) {
 //
 // In all the remaining cases B will overwrite A.
 func mergeMaps(dst interface{}, src interface{}) (interface{}, error) {
-	if dst == nil {
-		return src, nil
-	}
-
-	if src == nil {
-		return dst, nil
-	}
-
 	switch s := src.(type) {
 	case map[interface{}]interface{}:
+
+		if dst == nil {
+			dst = make(map[interface{}]interface{})
+		}
+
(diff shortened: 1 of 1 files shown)


akshayjshah:
Hi! Thanks so much for this fix. Can you see if this issue is still reproducible on the head of the master branch? There were a bunch of bugs like this, so we ended up completely rewriting and simplifying the package internals.

I *hope* that we've fixed your bug. If we haven't, we'll need to start working on a bugfix from scratch.

dgladkov:
The issueI had is fixed in master, however my project specific use case prevents from upgrading past 1.1.0 as we have a custom Provider with dynamic key lookup, so we're relying on the fact that merge happens on lookup, not on init. I couldn't reimplement that behavior using the new project design.

akshayjshah:
I see. We're not planning to keep dynamic providers - the code on master is the direction we're taking with this project.

If your environment relies on dynamic providers, this library won't be a good fit for you going forward. You're better off forking the `v1.1.0` code, trimming it down to just the bits you use, and treating it as a standalone project.

dgladkov:
I see, thanks! Given that this PR fixes a known issue, do think it's worth cutting a 1.1.1 bugfix release? Otherwise feel free to close.

akshayjshah:
I'd prefer not to patch this in, mostly because this is only one of a very large number of known bugs in the `v1.1.0` release. If you end up forking this package, I'll happily direct anyone using a dynamic provider to use your fork.

PR #91, opened 2018-07-02
Prepare for v1.2.0

akshayjshah:
Update changelog to v1.2.0. The version constant is already at the
correct value.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-## Unreleased
+## v1.2.0 - 2018-06-28
 ### Added
 - Add `NewYAML`, a new `Provider` constructor that lets callers mix Go values,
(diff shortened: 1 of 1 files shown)


dgladkov:
Consider making a 2.0.0 release instead.
* Dynamic providers are no longer supported.
* `NewValue` behavior is completely changed - it used to just build a `Value` struct, however now it's actually looking up a value from Provider. As custom `Provider.Get` implementations return a `Value` constructed from `NewValue`, this will lead to infinite recursion in most practical use cases.

Because public API was redesigned and old deprecated API is largely backwards incompatible, I feel that 2.0 release is justified. This will also give you opportunity to remove `NewProviderGroup` and and `NewValue`.

akshayjshah:
I hear you. Bugfixes are always changes in behavior, and there were so many bugs in this package that fixing them changed a lot of behavior. The impact of these changes on static YAML providers is minimal, but your dynamic providers will need quite a bit more work to update. (As discussed on your bugfix PR, I encourage you to fork this package if you prefer the current APIs and behavior.)

I'd love to delete the deprecated APIs and release this as a 2.0, but it's more important to me to keep compatibility with the many users of this code pinned to `^1`. It's unfortunate that we cut a 1.0 release with so many bugs and warty APIs :/

PR #92, opened 2018-07-05
[WIP] Handle empty sources correctly

akshayjshah:
When reading empty files (or using otherwise empty sources), we need to
handle `io.EOF` correctly. Special-case end-of-file where relevant, and
document why we don't need to handle it elsewhere.


akshayjshah:
Rethinking the approach here.

PR #95, opened 2018-07-10
NewProviderGroup: Handle empty sources correctly

abhinav:
We were encountering a bug where `NewProviderGroup` was interpreting
empty files correctly despite the previous fix. Specifically, the
following works with `NewYAML`,

    # foo.yaml contains:
    #
    #   foo:
    #     bar: 42

    p := NewYAML(File("foo.yaml"), File("empty.yaml")))
    fmt.Println(p.Get("foo.bar").Value())
    // Output: 42

But the following does not,

    p1 := NewYAML(File("foo.yaml"))
    p2 := NewYAML(File("empty.yaml"))
    p  := NewProviderGroup(p1, p2)
    fmt.Println(p.Get("foo.bar").Value())
    // Output: nil

It doesn't work because `NewProviderGroup` works by calling `Get(Root)`
on the providers and building a new `Provider` with the merged contents.
`p2` doesn't know the difference between empty file and `nil` so it
returns `nil` for its contents, which the merging logic interprets as
unsetting everything at that level.

Unfortunately, switching this call site from `NewProviderGroup` to
`NewYAML` isn't possible because `NewYAML` applies options like
`Permissive` and `Expand` to all files and we don't want to `Expand` on
one of the files.

To fix this, I made two changes:

- YAML providers now have a record of whether they were empty or not.
  When a `Get` is performed on a`YAML` provider, the resulting `Value`
  now accurately knows whether it has a value.
- `NewProviderGroup` includes the contents of a provider only if its
  `Value.HasValue` returns true.

I added failing test cases for this case. I replicated that test for the
null sources case to avoid accidentally breaking it in the future.

### DIFF
--- a/config_ext_test.go
+++ b/config_ext_test.go
@@ -649,36 +649,59 @@ func TestEmptySources(t *testing.T) {
 	empty := ""
 	comment := "# just a comment"
 
-	tests := []struct {
-		desc    string
-		sources []string
-		expect  interface{}
-	}{
-		{"no sources", []string{}, nil},
-		{"empty base", []string{empty, full}, val},
-		{"empty override", []string{full, empty}, val},
-		{"empty base and override", []string{empty, empty}, nil},
-		{"comment-only base", []string{comment, full}, val},
-		{"comment-only override", []string{full, comment}, val},
-		{"empty base and comment-only override", []string{empty, comment}, nil},
-		{"comment-only base and empty override", []string{comment, empty}, nil},
+	runTests := func(t *testing.T, newProvider func([]string) Provider) {
+		tests := []struct {
+			desc    string
+			sources []string
+			expect  interface{}
+		}{
+			{"no sources", []string{}, nil},
+			{"empty base", []string{empty, full}, val},
+			{"empty override", []string{full, empty}, val},
+			{"empty base and override", []string{empty, empty}, nil},
+			{"comment-only base", []string{comment, full}, val},
+			{"comment-only override", []string{full, comment}, val},
+			{"empty base and comment-only override", []string{empty, comment}, nil},
+			{"comment-only base and empty override", []string{comment, empty}, nil},
+		}
+
+		for _, tt := range tests {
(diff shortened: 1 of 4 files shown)


constructors.go
```diff
@@ -58,7 +58,10 @@ func NewProviderGroup(name string, providers ...Provider) (Provider, error) {
 	opts := make([]YAMLOption, 0, len(providers)+2)
 	opts = append(opts, Name(name), Permissive())
 	for _, p := range providers {
-		opts = append(opts, Static(p.Get(Root).Value()))
+		v := p.Get(Root)
+		if v.HasValue() {
```
zmt: The "magic".

config_ext_test.go
```diff
@@ -649,36 +649,59 @@ func TestEmptySources(t *testing.T) {
...
+				var err error
+				providers[i], err = NewYAML(Source(strings.NewReader(sources[i])))
+				require.NoError(t, err, "failed to create provider")
+			}
+			p, err := NewProviderGroup("foo", providers...)
+			require.NoError(t, err, "failed to create provider")
+			return p
 		})
-	}
+	})
```
akshayjshah: Nice, I like this a lot.

constructors.go
```diff
@@ -58,7 +58,9 @@ func NewProviderGroup(name string, providers ...Provider) (Provider, error) {
 	opts := make([]YAMLOption, 0, len(providers)+2)
 	opts = append(opts, Name(name), Permissive())
 	for _, p := range providers {
-		opts = append(opts, Static(p.Get(Root).Value()))
+		if v := p.Get(Root); v.HasValue() {
+			opts = append(opts, Static(v.Value()))
+		}
```
akshayjshah: Since we don't have a reasonable way to pull it out of our internal code, we should un-deprecate this constructor. Let's remove the `Deprecated:` bit but keep a comment suggesting that people use `NewYAML` where possible.

PR #103, opened 2018-10-08
Re-expand env vars in WithDefault

akshayjshah:
To make `WithDefault` handle explicit nils correctly, we need to work
directly with the original YAML sources. When re-merging these sources
with the new default source, we neglected to re-expand environment
variables. This effectively made any Values created by `WithDefault`
behave as though environment variable expansion were disabled.

In this commit, we fix this bug by re-expanding env vars in
`WithDefault`. This isn't ideal, since it requires reaching out into the
environment again, but it's very simple. The more-correct alternative
requires parsing all environment variables out of every YAML source and
looking them up during provider construction, then using that lookup
table in subsequent calls to `WithDefault`. Since `WithDefault` is
deprecated and the environment doesn't usually change over the life of
the process, sticking with the simpler implementation seems like a
pragmatic tradeoff.

### DIFF
--- a/config.go
+++ b/config.go
@@ -24,13 +24,11 @@ import (
 	"bytes"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"strings"
 
 	"go.uber.org/config/internal/merge"
 	"go.uber.org/config/internal/unreachable"
 
-	"golang.org/x/text/transform"
 	yaml "gopkg.in/yaml.v2"
 )
 
@@ -50,6 +48,7 @@ const _separator = "."
 type YAML struct {
 	name     string
 	raw      [][]byte
+	lookup   LookupFunc // see withDefault
 	contents interface{}
 	strict   bool
 	empty    bool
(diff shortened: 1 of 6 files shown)


akshayjshah:
Good suggestion! Done in https://github.com/uber-go/config/pull/103/commits/45e338c1680ba210665ede4a36b919ada29d28e5.

config.go
```diff
@@ -219,9 +217,19 @@ func (y *YAML) withDefault(d interface{}) (*YAML, error) {
 		return nil, fmt.Errorf("merging default and existing YAML failed: %v", err)
 	}
 
+	// Since we've gone back to working with the original input sources, we also
+	// need to re-expand environment variables. Since we only need to do this to
+	// support this deprecated method, keep things simple and reach out into the
+	// environment again.
+	merged, err = expandVariables(y.lookup, merged)
```
prashantv: this looks fine, but a lot of this logic looks very similar to `NewYAML`. Can we instead use `NewYAML` (maybe we provide an unexported option that lets you add `[]config.source` directly). I think the code could instead be:

```
opts := []yaml.YAMLOption{
  Name(y.Name),
  Expand(y.lookup),
  Source(rawDefault),
  rawSources(y.raw),
}
if !y.strict {
  opts = append(opts, Permissive())
}

return NewYAML(opts...)
```

The "merging default and existing YAML" error message might not be as specific to default, but it seems like less code to maintain, and less likely to run into problems like this.

abhinav: done

akshayjshah:
Thanks for pushing the cleanup - looks great!

PR #110, opened 2019-10-23
Ensure replacement values are properly quoted

### LINKED ISSUE #109: Ensure replacement values are properly quoted when necessary
YAML requires that strings which contain characters that conflict with YAML syntax elements must be quoted in order to be parsed properly. For example, YAML does not allow unquoted strings to begin with the `@` character so any string that begins with this character must be quoted. The `expandTransformer`, however, does not currently check if a replacement value needs to be quoted and will produce invalid YAML if it's lookup function returns a value that must be quoted. This problem is reproduced in the following program:

```go
package main

import (
	"fmt"
	"os"
	"strings"

	"go.uber.org/config"
)

func main() {
	environment := map[string]string{"FOO": "@foo"}
	lookup := func(key string) (string, bool) {
		s, ok := environment[key]
		return s, ok
	}

	yaml := strings.NewReader("key: ${FOO}")
	p, err := config.NewYAML(config.Source(yaml), config.Expand(lookup))
	if err != nil {
		fmt.Printf("Error: Unable to construct YAML provider: %v.\n", err)
		os.Exit(1)
	}
	fmt.Println(p.Get("key").Value())
}
```

The output of this program is:

```text
Error: Unable to construct YAML provider: couldn't decode merged YAML: yaml: found character that cannot start any token.
```

The issue is that the YAML template

```yaml
key: ${FOO}
```

is converted to the following YAML source

```yaml
key: @foo
```

but in order for this source to be valid YAML it should be

```yaml
key: "@foo"
```

jeromefroe:
Fixes #109

### DIFF
--- a/expand.go
+++ b/expand.go
@@ -27,6 +27,7 @@ import (
 	"strings"
 
 	"golang.org/x/text/transform"
+	yaml "gopkg.in/yaml.v2"
 )
 
 const (
@@ -68,7 +69,11 @@ func replace(lookUp LookupFunc) func(in string) (string, error) {
 		}
 
 		if envVal, ok := lookUp(key); ok {
-			return envVal, nil
+			buf, err := yaml.Marshal(envVal)
+			if err != nil {
+				return "", fmt.Errorf("unable to marshal value of %q to YAML: %v", key, err)
+			}
+			return string(buf), nil
 		}
 
 		if def == "" {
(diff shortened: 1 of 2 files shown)


jeromefroe:
As @prashantv pointed out in #109 this approach will result in integer values being quoted also which isn't what we want. I'll look into the problem further to see if there is a solution we can find for both integers and strings that must be quoted.

jeromefroe:
This approach doesn't work and I was able to change the value in question in our case such that the new value doesn't need to be quoted.

PR #111, opened 2019-11-05
Switch to Go modules

abhinav:
This switches the library to Go modules, simplifying the Makefile and Travis
setup.

I also set up linting while at it, and fixed the issues it found.


.gitignore
```diff
@@ -11,3 +11,6 @@ coverage.txt
 *.pprof
 /.bin
 /.cache
+/bin
```
SyrieBianco: Is there a reason we've got a `/bin` and a `/.bin`? Might make sense to combine in a later clean up

abhinav: /bin because we use it in the Makefile now. `/.bin` probably a leftover from old copy-pasta. Will make an attempt to clean up gitignores in a future change.

